PR #509, opened 2024-03-09
New stdlib math methods

dorian3343:
Added subtraction and multiplication. Check if the way I added it was correct since I used a bool flag for the first port (subtracting from 0 and multiplying from 0 would give bad answers)


internal/runtime/funcs/int_subtracter.go
```diff
@@ -0,0 +1,47 @@
+package funcs
+
+import (
+	"context"
+	"github.com/nevalang/neva/internal/runtime"
+)
+
+type intSubtracter struct{}
```
emil14, line 8: isn't subtract**o**r the correct word?

std/builtin/base.neva
```diff
@@ -25,6 +25,12 @@ component {
 #extern(int int_adder, float float_adder, string string_adder)
 pub component Adder<T int | float | string>(stream maybe<T>) (res T)
 
+#extern(int int_subtracter, float float_subtracter, string string_subtracter)
+pub component Subtracter<T int | float | string>(stream maybe<T>) (res T)
+
+#extern(int int_multiplier, float float_multiplier, string string_multiplier)
+pub component Multiplier<T int | float | string>(stream maybe<T>) (res T)
```
emil14, line 32: How are you gonna multiply strings?

std/builtin/base.neva
```diff
@@ -25,6 +25,12 @@ component {
 #extern(int int_adder, float float_adder, string string_adder)
 pub component Adder<T int | float | string>(stream maybe<T>) (res T)
 
+#extern(int int_subtracter, float float_subtracter, string string_subtracter)
+pub component Subtracter<T int | float | string>(stream maybe<T>) (res T)
```
emil14, line 29: How are you gonna substract strings?

internal/runtime/funcs/int_multipier.go
```diff
@@ -0,0 +1,48 @@
...
+		return nil, err
+	}
+
+	resOut, err := io.Out.Port("res")
+	if err != nil {
+		return nil, err
+	}
+
+	return func(ctx context.Context) {
+		flag := false
```
emil14, line 23: u don't need this, just use `1` as initial value

internal/runtime/funcs/int_subtracter.go
```diff
@@ -0,0 +1,47 @@
...
+			select {
+			case <-ctx.Done():
+				return
+			case streamItem := <-streamIn:
+				if streamItem == nil {
+					select {
+					case <-ctx.Done():
+						return
+					case resOut <- runtime.NewIntMsg(res):
+						res = 0
```
emil14, line 34: u need to set flag back to false

internal/runtime/funcs/int_subtracter.go
```diff
@@ -0,0 +1,47 @@
...
+						continue
+					}
+				}
+				if !flag {
```
emil14, line 38: nit

How about using `var res *int64` and have 1 variable instead of 2? Only if that will make code simpler

internal/runtime/funcs/int_multipier.go
```diff
@@ -0,0 +1,48 @@
...
+			select {
+			case <-ctx.Done():
+				return
+			case streamItem := <-streamIn:
+				if streamItem == nil {
+					select {
+					case <-ctx.Done():
+						return
+					case resOut <- runtime.NewIntMsg(res):
+						res = 0
```
emil14, line 35: set to 1 instead of 0

emil14:
Please add some example(s) and (optional) e2e test (I'll add it myself in case it's hard. No worries)

internal/runtime/funcs/int_subtracter.go
dorian3343: This will complicate the code for no benefit IMO

PR #515, opened 2024-03-13
Added list length component

dorian3343:
Added a Length component to the stdlib which checks the length of a list and returns it. Check the type signature because i tried doing `List<T> ` as inport and it didn't work so for now this is the type signature.

```
pub component Length<T>(data T) (res int)
```


e2e/tests/list_length/main/main.neva
```diff
@@ -0,0 +1,12 @@
+const n any = [4,123]
```
emil14, line 1: `const n list<int> = [4, 123]`

e2e/tests/list_length/main/main.neva
```diff
@@ -0,0 +1,12 @@
...
+component Main(start any) (stop any) {
+	nodes {
+	    printer Printer<any>
+	    len Length<any>
```
emil14, line 5: Let's call component itself `Len`, just like in Go

examples/13_array_length/main.neva
```diff
@@ -0,0 +1,12 @@
+const n any = [4,123]
```
emil14, line 1: If there's no difference between example and file in `e2e/tests`, remove file in tests and simply use example for e2e test

internal/runtime/funcs/length.go
```diff
@@ -0,0 +1,35 @@
+package funcs
+
+import (
+	"context"
+	"github.com/nevalang/neva/internal/runtime"
+)
+
+type length struct{}
```
emil14, line 8: Let's call it `listLen` and use only for lists at runtime level

std/builtin/base.neva
```diff
@@ -49,4 +49,11 @@ type EqElse<T any> struct {
 }
 
 #extern(int int_eq, float float_eq, string string_eq)
-pub component Eq<T int |float | string>(a T, b T) (then T, else EqElse<T>)
\ No newline at end of file
+pub component Eq<T int |float | string>(a T, b T) (then T, else EqElse<T>)
+
+// List
+#extern(length)
+pub component Length<T>(data T) (res int)
```
emil14, line 56: Let's use overloading to support maps (you don't have to write runtime function, just the interface here)

```neva
#extern(list list_len, map map_len)
pub component Len<T list<any> | map<any>>(data T) (res int)
```

internal/runtime/funcs/registry.go
```diff
@@ -31,5 +31,7 @@ func CreatorRegistry() map[string]runtime.FuncCreator {
 		"int_parser": intParser{},
 		// regexp
 		"regexp_submatcher": regexpSubmatcher{},
+		//array
```
emil14, line 34: please use Nevalang terminology, it's "list", not array

e2e/tests/list_length/main/main.neva
emil14, line 1: do we need this file?

PR #523, opened 2024-03-15
Basic array indexing 

dorian3343:
Need to add an e2e test but wanted to get your thought's


std/builtin/base.neva
```diff
@@ -55,5 +55,5 @@ pub component Eq<T int |float | string>(a T, b T) (then T, else EqElse<T>)
 #extern(list list_len, map map_len)
 pub component Len<T list<any> | map<any>>(data T) (res int)
 
-
-
+#extern(get_index)
```
emil14, line 58: Let's use the same naming `#extern(index)` or maybe `list_index` just to be consistent with the rest of the stdlib.

Proper naming is something we still need to figure out but there is no named like that. Probably more idiomatic naming would be `indexer` but I suggest stick with `index` for now.

internal/runtime/funcs/index.go
```diff
@@ -0,0 +1,53 @@
...
+			case <-ctx.Done():
+				return
+			case data := <-listIn:
+				select {
+				case <-ctx.Done():
+				case idx := <-indexIn:
+					select {
+					case <-ctx.Done():
+					default:
+						if idx.Int() < 0 || idx.Int() >= int64(len(data.List())) {
```
emil14, line 43: nit

I suggest adding a couple of variables just for readability and so we don't call same functions twice.

```
intIdx := idx.Int()
lst := data.List()
```

examples/14_list_index/main.neva
```diff
@@ -0,0 +1,18 @@
+const n list<int> = [1,1,5,112,69,420]
+const j int = 4
+component Main(start any) (stop any) {
+	nodes {
+        index Index<any>
+        printer Printer<any>
+	}
+	net {
+		:start -> ($n -> index:data,$j -> index:idx)
```
emil14, line 9: nit: formatting

examples/14_list_index/e2e_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"os"
+	"os/exec"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestInterfacesSimple(t *testing.T) {
```
emil14, line 11: please rename to just `Test`

examples/14_list_index/e2e_test.go
```diff
@@ -0,0 +1,30 @@
...
+	err := os.Chdir("../..")
```
emil14, line 12: I think this line is the problem, you probably need to `..` instead of `../..`

emil14, line 12:
```suggestion
	err := os.Chdir("..")
```

examples/14_list_index/e2e_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+func TestInterfacesSimple(t *testing.T) {
+	err := os.Chdir("..")
+	require.NoError(t, err)
+
+	wd, err := os.Getwd()
+	require.NoError(t, err)
+	defer os.Chdir(wd)
+
+	cmd := exec.Command("neva", "run", "3_list_index")
```
emil14, line 19:
```suggestion
	cmd := exec.Command("neva", "run", "3_list_index")
```

PR #543, opened 2024-04-06
Implemented Iter

dorian3343:


std/builtin/base.neva
```diff
@@ -74,4 +76,12 @@ pub component Join(data list<string>) (res string)
 
 // String
 #extern(split)
-pub component Split(data string) (res list<string>)
\ No newline at end of file
+pub component Split(data string) (res list<string>)
+
+#extern(upper)
+pub component ToUpper(data string) (res string)
```
emil14, line 82: I guess these two are some kind of draft? Could you please remove them until you don't have actual implementation and working example?

examples/iter/main.neva
```diff
@@ -0,0 +1,19 @@
+const l list<int> = [50, 30, 20, 100]
```
emil14, line 1: nit: could you please name this consistently with other examples like e.g. `19_list_iteration`

emil14:
Also please increment minor version to `0.16.0` in `pkg/version.go`

PR #547, opened 2024-04-08
io: Introduce FileReader and FileWriter API

Catya3:
`FileReader` and `FileWriter` work for simple File IO. 

* Added working examples


examples/19_file_reader/main.neva
```diff
@@ -0,0 +1,16 @@
+import { io }
+
+const filename string = '19_file_reader/main.neva'
+
+component Main(start any) (stop any) {
+  nodes {
+    io.FileReader
```
emil14, line 7: how about naming it `ReadAll` like in Go since we're planning to add more flexible version in the future?

Also, I suggest using verbs e.g. ReadFile instead of FileReader (naming convention was changed in v0.16)

examples/20_file_writer/main.neva
```diff
@@ -0,0 +1,17 @@
+import { io }
+
+const {
+  filename string = 'file_writer_example.txt'
```
emil14, line 4: What will happen if file doesn't exist? will it throw error or create that file?

internal/runtime/funcs/file_reader.go
```diff
@@ -0,0 +1,57 @@
...
+		select {
+		case <-ctx.Done():
+			return
+		case name := <-filename:
+			f, err := os.Open(name.Str())
+			if err != nil {
+				select {
+				case <-ctx.Done():
+					return
+				case errPort <- runtime.NewStrMsg(err.Error()):
```
emil14, line 39: Could you please implement error like this

```go
runtime.NewMapMsg(map[string]runtime.Msg{
	"text": runtime.NewStrMsg(err.Error()),
})
```

Because errors in Nevalang are structures, not strings (they should be "wrappable")

I saw how error is implemented in `Index` it's actually not correct, that's my fault

internal/runtime/funcs/file_reader.go
```diff
@@ -0,0 +1,57 @@
...
+					return
```
emil14, line 40: I don't think we should return after error, that will lead to component instance termination. We probably need to `continue` for loop to receive next message

internal/runtime/funcs/file_reader.go
```diff
@@ -0,0 +1,57 @@
...
+	errPort, err := rio.Out.Port("err")
+	if err != nil {
+		return nil, err
+	}
+
+	return func(ctx context.Context) {
+		select {
+		case <-ctx.Done():
+			return
+		case name := <-filename:
```
emil14, line 33: your implementation will only handle one message, every runtime function should have for loop to handle infinite number of messages, checkout other runtime functions for inspiration

internal/runtime/funcs/file_reader.go
```diff
@@ -0,0 +1,57 @@
...
+			f, err := os.Open(name.Str())
+			if err != nil {
+				select {
+				case <-ctx.Done():
+					return
+				case errPort <- runtime.NewStrMsg(err.Error()):
+					return
+				}
+			}
+			defer f.Close()
```
emil14, line 43: I think we shouldn't defer file closing and instead close file right after each iteration. We receive message, then we handle it (read file to memory), then we close file and send data downstream

internal/runtime/funcs/file_reader.go
```diff
@@ -0,0 +1,57 @@
...
+			data, err := io.ReadAll(f)
+			if err != nil {
+				select {
+				case <-ctx.Done():
```
emil14, line 47: looks like `return` is missing

internal/runtime/funcs/file_writer.go
```diff
@@ -0,0 +1,60 @@
...
+			case <-ctx.Done():
+				return
+			case errPort <- runtime.NewStrMsg(err.Error()):
+				return
+			}
+		}
+		select {
+		case <-ctx.Done():
+			return
+		case sig <- runtime.EmptyMsg{}:
```
emil14, line 56: That wasn't obvious but you should send `nil`

internal/runtime/funcs/file_writer.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"context"
+	"os"
+
+	"github.com/nevalang/neva/internal/runtime"
+)
+
+type fileWriter struct{}
+
+func (c fileWriter) Create(rio runtime.FuncIO, msg runtime.Msg) (func(ctx context.Context), error) {
```
emil14, line 12: basically all the comments from fileReader are related here too

PR #559, opened 2024-04-19
Runtime: Add Decode, Encode for native Msg types

Catya3:
This makes the `runtime.Msg` native types easily embeddable in other structs so we can create compound struct types more easily.

Still TBD how/if to make `Encode` and `Decode` on structs more economical.

Example usage:

```go

type rgbaMsg struct {
	r runtime.IntMsg
	g runtime.IntMsg
	b runtime.IntMsg
	a runtime.IntMsg
}

func (c *rgbaMsg) Reset() { *c = rgbaMsg{} }
func (c *rgbaMsg) Decode(msg runtime.Msg) bool {
	c.reset()
	if msg == nil {
		return false
	}
	m := msg.Map()
	if m == nil {
		return false
	}
	return c.r.Decode(m["r"]) &&
		c.r.Decode(m["g"]) &&
		c.r.Decode(m["b"]) &&
		c.r.Decode(m["a"])
}
func (c *rgbaMsg) Encode() runtime.Msg {
	return runtime.NewMapMsg(map[string]runtime.Msg{
		"r": c.r,
		"g": c.g,
		"b": c.b,
		"a": c.a,
	})
}
```

For now this helps with internal dev and making runtime funcs. In the future, such code can be generated from the stdlib or user code libraries facilitating interop with Go.

### DIFF
--- a/internal/runtime/msg.go
+++ b/internal/runtime/msg.go
@@ -53,6 +53,16 @@ type IntMsg struct {
 func (msg IntMsg) Type() MsgType  { return IntMsgType }
 func (msg IntMsg) Int() int64     { return msg.v }
 func (msg IntMsg) String() string { return strconv.Itoa(int(msg.v)) }
+func (i *IntMsg) Reset()          { i.v = 0 }
+func (i *IntMsg) Decode(msg Msg) bool {
+	i.Reset()
+	if msg == nil || msg.Type() != IntMsgType {
+		return false
+	}
+	i.v = msg.Int()
+	return true
+}
+func (msg IntMsg) Encode() Msg { return msg }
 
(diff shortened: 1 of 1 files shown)


emil14:
I don't understand yet why we need this, would love to discuss in person :)

Generally my vision is this - message is what it is, not less or more. If some runtime function needs some private type then it should just be private type for that function inside `funcs` package. Runtime should not know anything about that.

Also this implementation is not thread-safe and not immutable.

But again, I might be missing something so would love to discuss :)

emil14:
We discussed in person that this is not needed because the runtime is unsafe which means it doesn't do validation. Compiler guarantees that program is correct

PR #570, opened 2024-04-26
fix(cli): use module root to correctly identify main pkg name (path)

emil14:


emil14:
Related to https://github.com/nevalang/neva/issues/571

PR #573, opened 2024-04-27
WIP: Feature std image

emil14:


examples/image/draw_over/main.neva
```diff
@@ -0,0 +1,33 @@
+import {
+  io
+  image
+}
+
+component Main(start) (stop) {
+  nodes {
+    // bgReader io.ReadAll,
```
emil14, line 8: remove comments or example itself

examples/image/minimal_png/main.neva
```diff
@@ -0,0 +1,23 @@
+import {
+  io
+  image
```
emil14, line 3: TODO discuss image API

internal/runtime/funcs/image.go
```diff
@@ -0,0 +1,137 @@
+package funcs
+
+import (
+	"image"
+	"image/color"
+
+	"github.com/nevalang/neva/internal/runtime"
+)
+
+func decodeInt(i *int64, m runtime.Msg) bool {
```
emil14, line 10: am I right that we'll remove these?

internal/runtime/funcs/image_decode.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"context"
+	"image"
+	"image/png"
+	"strings"
+
+	"github.com/nevalang/neva/internal/runtime"
+)
+
+type imageDecode struct{}
```
emil14, line 12: I'm not sure if all of these are used in examples but in case they're not, I suggest only add those runtime funcs that are actually used

examples/image/draw_over/main.neva
Catya3: Decided to remove the example as `Draw` is going away

internal/runtime/funcs/image.go
Catya3: Removed Decode for primitive types

internal/runtime/funcs/image_decode.go
Catya3: I need an example for read/write on image

Catya3: Removed for now...

Catya3:
It looks like we might be getting issues from the use of `Struct` builders in the example code:
> std@0.20.0/builtin/core.neva:0:0 can't resolve argument: can't get type def from scope by ref: import not found: image

Catya3:
The latest error is: `std@0.22.0/builtin/core.neva:0:0 can't resolve struct field: color: can't get type def from scope by ref: entity not found: 'RGBA'`

PR #580, opened 2024-04-30
Stream to List

emil14:


emil14:
In this PR I might

1. Rename `map` to `dict` to avoid confusion with `Map` component and to support naming convention for components `List` and `Dict` that can receive stream and produce `list` and `dict`. BTW I'm not sure about the naming because we also need a way to build these _collections_ from array-port slots. What the naming should be? What is the most common use-case?
2. Component that turns array-port into stream could be named simply `Stream`, how about that?
3. My first intention was to implement `Map/Filter/Reduce` and maybe something like `For/ForEach/Loop` (for side-effects)
4. I might end up with fixing some of the existing issues to support everything what I need

PR #592, opened 2024-05-05
Runtime tracing for messages

emil14:


emil14:
Let the branch be here for a while but this implementation has some problems

1. We have to extend `Msg` interface
2. Messages that are inside other messages (map/struct/list/etc) won't be properly wrapped
3. It's actually not clear 100% whether we need wrapping like this (we probably should implement error handling first)

PR #607, opened 2024-05-09
Implemented new logic components

dorian3343:
Implemented two new logic components to improve DX.

### Assert component
**Type signature**
```
pub Assert<T>(val T, val2 T) (res bool)
```
* Note to review, please suggest better names for assert inports

Compares two values and  returns if they're equal

### If component
**Type signature**
```
pub If(val bool) (ok any, else any)
```

Takes in a boolean and starts one of the two branches. Ok if the bool is true, else if its false

### Example usage

```
component Main(start) (stop) {
    nodes { Println, assert Assert<string>, if If }
    :start -> [
        ('neva is cool' -> assert:val),
        ('neva' -> assert:val2)
       ]

       assert:res -> if

       if:ok -> ('They match' -> println -> :stop)
       if:else -> ('They do not match' -> println -> :stop)
}

```


emil14:
1. I suggest call it Eq because "Assert" sounds like it should return error or panic
2. I suggest have two outports else and then instead of sending boolean. Send boolean is first thing that comes to mind but then you should "assert" that boolean (is it true or false?) and implement routing. I figured out some time ago that sending booleans isn't much useful in Neva compared to having several outports
3. I would like to discuss possibility of having array-inport so we can compare N values instead of just two, also discuss if we need this for streams

dorian3343:
1. Renamed it as you wished.
2. I actually think sending a boolean is good, since you can build on top of the 'if' (we can add more and build an ecosystem for simple logic syntax). Having two outports forces you to use it in a specific way instead of it being flexible.
3.You can just use iter

emil14:
> Having two outports forces you to use it in a specific way instead of it being flexible

Why? Could you please provide some proof?

> `If` and `Eq` will pair nicely with Iter

1. I don't think user should use `Iter` most of the time preferring something like `map/filter/reduce/for`
2. Could you please show some examples?

PR #611, opened 2024-05-09
Lock: Add lock mechanism with array of input sigs

Catya3:
Name TBD

This is used when you have multiple `sig` and you want to wait on _all_ of them to do something. A very common scenario.

```
:start -> (0 -> polyLock:data)
0 -> println -> polyLock:sig[0]
1 -> println -> polyLock:sig[1]
2 -> println -> polyLock:sig[2]
polyLock -> :stop
```

### DIFF
--- a/internal/runtime/funcs/lockall.go
+++ b/internal/runtime/funcs/lockall.go
@@ -1,0 +1,61 @@
+package funcs
+
+import (
+	"context"
+	"errors"
+
+	"github.com/nevalang/neva/internal/runtime"
+)
+
+type lockAll struct{}
+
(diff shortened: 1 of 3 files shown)


emil14:
> A very common scenario.

If that's the case, please add some examples :)

emil14:
We decided to create `sync` package

emil14:
BTW other names could be `LockAll` or `LockMany`

emil14:
@Catya3 I'm going to close this one because it has conflicts and we don't have need for this component at the moment. We can reopen this at any moment

Issue #575, opened 2024-04-28
No easy way of predictable termination when working with streams (sequences)


emil14:
# Problem

Let's take a simplest problem - "print each number in a stream". Turns out current implementation doesn't allow one to do so without introducing some extra complexity.

# Demonstration

Here's naive attempt to program this logic:

```neva
$lst -> iter -> unwrap
unwrap:some -> println -> del
unwrap:none -> :stop
```

The problem is with this line `unwrap:none -> :stop`.

There's no guarantee that `unwrap:none` is sending _after_ all elements from `unwrap:some` are printed. In fact, there's higher chance that we terminate _before_ we print all the elements because printing takes time.

The root of this problem is **parallel processing without synchronization**. If only we could somehow _await_ for all elements to be printed.

# Bad solution

One way to fix this is to get rid of parallelism and introduce a pipeline. If data moves through the single flow then there's no asynchronism and everything happens in sequential manner.

Let's print messages before unwrapping them!

```neva
$lst -> iter -> println -> unwrap
unwrap:none -> :stop
```

Because every message that comes to `-> unwrap` is already printed there's a guarantee that when `unwrap:some ->` is sending, all values are processed (printed in this case).

However, there's a problem with this solution. Let's say our `$lst` is `[1,2,3]`. The output will be not `1\n2\n3\n`, it will be `1\n2\n3\n<nil>\n`! It could be even worse if we change how `maybe` is represented as a string in the future. In other words - this is not a solution because _program operates in a different way_.

---

Important to understand that the problem is not about `maybe`, it's about parallelism and lack of synchronization. However, changing how stream works OR maybe how Iter works _could_ fix this. As well as changing something else.

emil14:
# Use structs with indexes instead of `maybe`

Imagine each stream's element is wrapped into something like this

```
pub type StreamItem<T> struct {
  data T
  idx int
}
```

Component that emits stream uses something like this `StreamItem | StreamDelimiter`

`StreamDelimiter` could be `{idx int}` or even just `int`

# Lock

When `unwrap:none` (or it's alternative) is sending we `lock` until some "processor" is done with the message from `unwrap:some` (or it's alternative). The trick is to unlock _only if_ message from `:some` and `:none` has the same index (or if they are equal themselves, depending on the API)

emil14:
^ question is how we handle maps this way?

emil14:
When this will be implemented/fixed it might be possible to #568

emil14:
[...]

PR #612, opened 2024-05-10
Unwrap Component 2.0

dorian3343:
## A new and improved version of the original Unwrap

**Why?**
Before the new interface for streaming, neva's Iter was a simple and elegant language construct, however since v0.20.0 it has become ugly and quite unusable.

Before:
```const lst list<int> = [50, 30, 20, 100]

component Main(start) (stop) {
    nodes { Println<maybe<int>>, Iter<int>, Unwrap<int> }
    net {
        :start -> ($lst -> iter -> println -> unwrap)
        unwrap:none -> :stop
    }
}
```

After:
```
const lst list<int> = [50, 30, 20, 100]

component Main(start) (stop) {
    nodes { Println<stream<int>>, Iter<int>, Match<bool> }
    :start -> ($lst -> iter -> println) 
    println.last -> match:data
    true -> match:case[0] -> :stop
}
```

**Solution**
The solution was upgrading unwrap to a more modern implemnation which turns that same code into this.
```const lst list<int> = [50, 30, 20, 100]

component Main(start) (stop) {
    nodes { Println<int>, Iter<int>,Unwrap<int>,Del}
    :start -> ($lst -> iter -> unwrap)
    unwrap:some -> println -> del
    unwrap:none -> :stop
}

```

**Use case**

Heres how easy it is to print numbers from 1 too 99, now imagine how easy it becomes to rewrite the 99 bottles example
```
component Main(start) (stop) {
    nodes { Println<int> Unwrap<int>,Del,Range}
        :start -> [
            (1 -> range:from),
            (100 -> range:to)
        ]
    range -> unwrap
    unwrap:some -> println -> del
    unwrap:none -> :stop
}

```

### DIFF
--- a/examples/iterate_over_list/e2e_test.go
+++ b/examples/iterate_over_list/e2e_test.go
@@ -22,11 +22,7 @@ func Test(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(
 		t,
-		`{"data": 50, "idx": 0, "last": false}
-{"data": 30, "idx": 1, "last": false}
-{"data": 20, "idx": 2, "last": false}
-{"data": 100, "idx": 3, "last": true}
-`,
+		"50\n30\n20\n100\n",
 		string(out),
 	)
 
--- a/examples/iterate_over_list/main.neva
+++ b/examples/iterate_over_list/main.neva
@@ -1,8 +1,9 @@
 const lst list<int> = [50, 30, 20, 100]
 
 component Main(start) (stop) {
-    nodes { Println<stream<int>>, Iter<int>, Match<bool> }
-    :start -> ($lst -> iter -> println) 
-    println.last -> match:data
-    true -> match:case[0] -> :stop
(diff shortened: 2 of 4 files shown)


emil14:
I'll copy the essence from the Discord just for the history:

1. `unwrap` is for `maybe`, not stream
2. The need for a separate component that extracts the data from stream item is not clear because stream item is just a structure
3. By doing this changes we hoped it will help with #575 but actually it doesn't (I've provided example with `time.Sleep` that proved that program can _unpredictably_ terminate before all elements of stream were processed)

dorian3343:
What do you think about the name Extract? Although it doesn't fix #575 (it help's slightly, a temporary bandaid is better than an open wound metaphorically) it's still useful for not having the user have to manually access the data.

emil14:
After discussion on a Discord I decided to close this one. I know you've put effort into it and I appreciate it!

The main reason for closing is that we don't need a component to extract data from stream item because user normally shouldn't do it at all. That was attempt to fix termination problem but we need to go different direction.

PR #626, opened 2024-05-13
New logic std

dorian3343:
*moved logic from core to logic
*added logic gate operators (AND,NOT,OR) you can implement any gate with a NAND so now all are possible in pure neva
*implented IsGreater and IsEven
*made Range work in reverse also (it would go on forever before)
*added examples for all of these, modified copare values to showcase NOT


std/builtin/logic.neva
```diff
@@ -0,0 +1,26 @@
...
+pub component And(A bool, B bool) (res bool)
+
+#extern(or)
+pub component Or(A bool, B bool) (res bool)
+
+#extern(int int_is_greater, float float_is_greater)
+pub component IsGreater<T int | float>(actual T, compared T) (res bool)
+
+#extern(int_is_even)
+pub component IsEven(data int) (res bool)
```
emil14, line 23: 1. I don't think we need this in stdlib
2. It's possible to implement it in pure neva, no need for runtime function

std/builtin/logic.neva
```diff
@@ -0,0 +1,26 @@
...
+pub component Not(data bool) (res bool)
+
+#extern(and)
+pub component And(A bool, B bool) (res bool)
+
+#extern(or)
+pub component Or(A bool, B bool) (res bool)
+
+#extern(int int_is_greater, float float_is_greater)
+pub component IsGreater<T int | float>(actual T, compared T) (res bool)
```
emil14, line 20: 1. Let's call it just `Gt`
2. Should also be able to compare `strings` (you don't have to implement runtime function, just add to overloading)

dorian3343: 1. Sure

2. That doesnt match though? Youre not expecting it to be greater youre just checking

emil14: 1. Should be consistent with `Eq`. `Gt/Lt/Eq` are pretty common naming you can find in many places. You can also add comment that will explain that Gt stands for "Greater Than"
2. This is not how they are compared in almost any other language including Go. Try to compare two characters by their length
3. It could mean "expected to greater than this". Don't make user think, it will be easier with consistent naming. This also about "Greater vs Gt". User will see `Eq` and will expect it to be `Gt/Lt`, not Greater/Less

dorian3343:
I decided to remove the num_comparison example since IsEven got removed which made it pointless (we already have an example for If and Gt doesnt require one)

std/builtin/logic.neva
dorian3343: Closing -> Rempved IsEven

PR #628, opened 2024-05-14
Sync: Add WaitGroup for counting lock use cases

Catya3:
Add example and e2e test

## On using Go's `sync.WaitGroup` internally

`WaitGroup` panics when the count gets below 0 so we have to be careful to take the count first, then take all the signals.

This may lead to panics in arithmetic use cases I.e. when passing `len(a)-len(b) -> wg; wg -> :done`. This is input dependent on the lengths of a and b. User needs to guard against this case.

### DIFF
--- a/examples/wait_group/main.neva
+++ b/examples/wait_group/main.neva
@@ -1,12 +1,12 @@
 import { sync }
 
 component Main(start) (stop) {
-  nodes { wg sync.WaitGroup, Println }
+  nodes { sync.Wg, Println }
   :start -> [
-    (3 -> wg),
-    ('Hello' -> println -> (-1 -> wg)),
-    ('Neva' -> println -> (-1 -> wg)),
-    ('World!' -> println -> (-1 -> wg))
+    (3 -> wg:count),
+    ('Hello' -> println -> (true -> wg:sig)),
+    ('Neva' -> println -> (true -> wg:sig)),
+    ('World!' -> println -> (true -> wg:sig))
   ]
   wg -> :stop
 }
--- a/examples/wait_group/main.neva
+++ b/examples/wait_group/main.neva
@@ -1,0 +1,12 @@
+import { sync }
+
+component Main(start) (stop) {
+  nodes { wg sync.WaitGroup, Println }
+  :start -> [
+    (3 -> wg),
+    ('Hello' -> println -> (-1 -> wg)),
+    ('Neva' -> println -> (-1 -> wg)),
+    ('World!' -> println -> (-1 -> wg))
+  ]
+  wg -> :stop
+}
--- a/internal/runtime/funcs/wait_group.go
+++ b/internal/runtime/funcs/wait_group.go
@@ -10,7 +10,12 @@ import (
 type waitGroup struct{}
 
 func (g waitGroup) Create(io runtime.FuncIO, _ runtime.Msg) (func(ctx context.Context), error) {
-	addIn, err := io.In.Port("add")
+	countIn, err := io.In.Port("count")
+	if err != nil {
+		return nil, err
+	}
+
+	sigIn, err := io.In.Port("sig")
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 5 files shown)


internal/runtime/funcs/wait_group.go
```diff
@@ -0,0 +1,88 @@
...
+) func(ctx context.Context) {
+	return func(ctx context.Context) {
+		var (
+			mu    sync.RWMutex
+			ready bool  // ready when we have received a nonnegative count, protected by mu.
+			count int64 // internal count, protected by mu.
+		)
+
+		reset := func() {
+			mu.Lock()
```
emil14, line 38: why not just native wait group?

std/sync/sync.neva
```diff
@@ -0,0 +1,17 @@
...
+//  :start -> (2 -> wg:add)
+//  task1 -> (-1 -> wg:add)
+//  task2 -> (-1 -> wg:add)
+//  wg -> :stop  // Signals after both tasks are done.
+//
+// Example:
+//  nodes { wg WaitGroup }
+//  wg -> :stop  // Blocks forever.
+#extern(wait_group)
+pub component WaitGroup(add int) (sig any)
```
emil14, line 17: What do you think about this API?

```
WaitGroup(count int, sig any) (sig any)
```

It accepts `count` and then each `->sig` decrements by one

That's less flexible API but my personal experience with Go was that I was always `wg.Done()` instead of `wg.Add(-n)`

emil14:
1. wdyt about name `sync.Wg` and just putting in comment that wg stands for WaitGroup? It will allow to `nodes {sync.Wg}` without manual alias
2. I propose 3 alternative APIs because of 2 reasons 1) this one looks too flexible (I'm not sure about this, it's just my experience with wg in Go) and 2) I'm afraid about it might be unsafe. Here they are:

# Less Flexible And Safe

```
(count int, sig any) (sig any)
```

In Go we have `wg.Done()` that works exactly like this

# Safe and Flexible 1

```
(count int, add int) (sig any)
```

# Safe and Flexible 2

```
(count int, sub int) (sig any)
```

`sub` port works exactly like `add` but inverted, it accepts positive integers to subtract (because it looks like we gonna end up by always passing negative ints)

examples/wait_group/main.neva
```diff
@@ -0,0 +1,12 @@
...
+component Main(start) (stop) {
+  nodes { wg sync.WaitGroup, Println }
+  :start -> [
+    (3 -> wg),
+    ('Hello' -> println -> (-1 -> wg)),
+    ('Neva' -> println -> (-1 -> wg)),
+    ('World!' -> println -> (-1 -> wg))
+  ]
+  wg -> :stop
+}
```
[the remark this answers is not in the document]
Catya3: Read the description. It highly likely -1 reaches the wg before 3, which normally would cause the program to stop, but I had to add a protection to the component to guard against this.

Catya3:
You have not described how you expect `count` to work in your proposed example. Should `count` be accepted only once? Should it update the internal count? Should it add or subtract from the internal count?

Catya3:
I can rewrite the example to use Go's `sync.WaitGroup` internally, but know that `WaitGroup` panics if the count is less than 0. This makes it very difficult to use for the implementation in Neva because -1 can and often does reach the counter before N which will cause a panic. It is much easier to leave the current implementation.

std/sync/sync.neva
Catya3: I'm ok with adding a sig inport. 

Your proposed change also makes it impossible to add more numbers to the counter later on?

emil14:
> You have not described how you expect count to work in your proposed example. Should count be accepted only once? Should it update the internal count? Should it add or subtract from the internal count?

True

I was thinking about this semantics (simplified code), wdyt?

```go
c := <-count
for c>0 {
  <-sig
  c--
}
```

internal/runtime/funcs/wait_group.go
emil14: Let's continue both topics here

> I can rewrite the example to use Go's sync.WaitGroup internally, but know that WaitGroup panics if the count is less than 0. This makes it very difficult to use for the implementation in Neva because -1 can and often does reach the counter before N which will cause a panic. We will need a temporary Int64 to hold negative numbers to avoid panic. It is "much easier to leave the current implementation".

I might be wrong but to me "temporary Int64" feels like just having a variable. How is it "much easier to leave the current implementation"? However, I won't push. Any correct implementation is ok for me.

However, I hope you'll be ok if one day I'll try to reimplement it with `sync.WaitGroup` if I see that it's easier

emil14: I would also add that

> WaitGroup panics when the count gets below 0, since we may have multiple calls happening concurrently, we need to protect against this case.

To me feels like we just need to read from `count` before we read from `sig/add/sub` (second port). I might be wrong tho

std/sync/sync.neva
emil14: > impossible to add more numbers to the counter later on

Well if we're talking about single "iteration" then yes. You send count and bunch of singals, they work together as a single itarton.

After that however you're free to send new count, any int you want.

Do you feel like it's a problem? Did you have use-cases for this?

examples/wait_group/main.neva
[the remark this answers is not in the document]
emil14: > Read the description

Sorry, I'll be more careful

> It highly likely -1 reaches the wg before 3, which normally would cause the program to stop, but I had to add a protection to the component to guard against this.

Ok but I feel like this problem could be solved by having 2 ports

internal/runtime/funcs/wait_group.go
Catya3: There's no one implementation, I was trying to make something that would always accept a new int to count but I realize this might be very bad now because consts are sent in a for loop...? I'll stash this and rework it to make it more like your instance of count with signal import

Catya3:
It's done the signature is now `(count, sig) -> (sig)`

PR #637, opened 2024-05-16
Dependabot revert

emil14:


emil14:
no idea why this should help cuz no idea why dependabot could broke anything

PR #649, opened 2024-05-20
Std update

dorian3343:
* fixed Lt
* implemented Args component


std/io/io.neva
```diff
@@ -1,2 +1,5 @@
 #extern(scanln)
-pub component Scanln(sig any) (data string)
\ No newline at end of file
+pub component Scanln(sig any) (data string)
+
+#extern(args)
+pub component Args() (data list<string>)
```
emil14, line 5: We decided to have create `os` package to follow Go's std structure, don't we? :)

dorian3343: fixed

PR #654, opened 2024-05-23
[Idea]: Megadump of new components

Catya3:
FP and related components that will work together.

Please add comments and suggestions inline.

### DIFF
--- a/ideas/new-ops.txt
+++ b/ideas/new-ops.txt
@@ -1,0 +1,117 @@
+# Proposal for new stdlib operations
+
+This is a proposal for several APIs that could be added to stdlib.
+
+## `builtin` candidates
+
+Entity	Signature	Notes
+
+`Not`	`Not(x bool) (x bool)`	Return logical not operation applied to input `x`
+`Pair<A, B>`	`struct { first A; second B }`	A generic pair of items. Can stand in for other pair types we've introduced such as `ZipResult`
+`OrDefault<T>`	`OrDefault<T>(t T)`	Try to return `t` or `default` if `t` is unavailable. `default` can be injected with `IInjectable<T>`
+`Unzip<T, R>`	`Unzip<T, R>(seq stream<ZipResult<T, R>>) (first stream<T>, second stream<R>)`	The inverse of `Zip<T, R>` which is already in `builtin`
+
+## DI Interfaces
+
+Inject these no-inport interfaces rather than having it in the signature.
+
+If lambda components are ever supported, we can write this more nicely.
+
+```
+// Before
+rand.Bool
+// ...
+0.3 -> bool:p
+bool -> :out
+
+// After
+rand.Bool{() -> 0.3}
+// ...
+bool -> :out
+```
+
+There are multiple benefits to this pattern:
+
+* Allows more components to satisfy simpler interfaces, while still retaining configurability.
+* Moving configurability to the compilation step instead of runtime.
+* Remove ambiguity of ports that are only "skimmed" that is read from once at component initialization and then never again. i.e. any port that would be skimmed should be an injectable.
+
+Entity	Signature	Notes
+
+`IBool`	`() -> (bool)`	An interface for components that return a `bool`
+`IInt`	`() -> (int)`	An interface for components that return a `int`
+`IFloat`	`() -> (float)`	An interface for components that return a `float`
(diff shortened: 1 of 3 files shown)


Catya3:
# Addendum

## time

|Entity|Signature|Note|
|---|---|---|
|`Delay<T>`|`Delay(msg T) (msg T)`|delay message by an internal timer sec `IInt`|

ideas/new-ops.txt
```diff
@@ -0,0 +1,117 @@
+# Proposal for new stdlib operations
+
+This is a proposal for several APIs that could be added to stdlib.
+
+## `builtin` candidates
+
+Entity	Signature	Notes
+
+`Not`	`Not(x bool) (x bool)`	Return logical not operation applied to input `x`
```
emil14, line 9: We already have this in builtin/logic.neva

ideas/new-ops.txt
```diff
@@ -0,0 +1,117 @@
...
+`Pair<A, B>`	`struct { first A; second B }`	A generic pair of items. Can stand in for other pair types we've introduced such as `ZipResult`
```
emil14, line 10: 1. Do you feel like we gonna need this a lot?
2. If yes, - enough to be inside builtin?

ideas/new-ops.txt
```diff
@@ -0,0 +1,117 @@
...
+`OrDefault<T>`	`OrDefault<T>(t T)`	Try to return `t` or `default` if `t` is unavailable. `default` can be injected with `IInjectable<T>`
```
emil14, line 11: Use-cases are needed

ideas/new-ops.txt
```diff
@@ -0,0 +1,117 @@
...
+`Unzip<T, R>`	`Unzip<T, R>(seq stream<ZipResult<T, R>>) (first stream<T>, second stream<R>)`	The inverse of `Zip<T, R>` which is already in `builtin`
```
emil14, line 12: 1. Use-cases are needed
2. Would be cool to know if other languages have it

ideas/new-ops.txt
```diff
@@ -0,0 +1,117 @@
...
+
+## DI Interfaces
+
+Inject these no-inport interfaces rather than having it in the signature.
```
emil14, line 16: already discussed on Discord but for the history will copy here

any component must have at least 1 inport (and 1 outport) just so we can control it's execution

> without inport it's not possible to "trigger" it, it can only work like New - unconditionally in an infinite loop from the program start
> :sig -> (',' -> :out)
>
> doesn't have this problem

ideas/new-ops.txt
```diff
@@ -0,0 +1,117 @@
...
+
+```
+// Before
+rand.Bool
+// ...
+0.3 -> bool:p
+bool -> :out
+
+// After
+rand.Bool{() -> 0.3}
```
emil14, line 28: I'm not sure if I get this one. Why are use mixing integers and booleans? I know it's just an example but would be nice to understand it.

ideas/new-ops.txt
```diff
@@ -0,0 +1,117 @@
...
+// ...
+bool -> :out
+```
+
+There are multiple benefits to this pattern:
+
+* Allows more components to satisfy simpler interfaces, while still retaining configurability.
```
emil14, line 35: why?

ideas/new-ops.txt
```diff
@@ -0,0 +1,117 @@
...
+* Moving configurability to the compilation step instead of runtime.
```
emil14, line 36: why?

ideas/new-ops.txt
Catya3: `p` is the probability of true in this case. `(1-p)` Of false

ideas/new-ops.txt
```diff
@@ -0,0 +1,117 @@
...
+* Remove ambiguity of ports that are only "skimmed" that is read from once at component initialization and then never again. i.e. any port that would be skimmed should be an injectable.
```
emil14, line 37: Why do you think there is ambiguity?

emil14:
@Catya3 I'm going to close this one. I suggest to you github issues for this kind of thing. Thank you very much for your effort. I will go back to this PR for inspiration each time there will be need for this. For now let's stick for existing usecases

Issue #644, opened 2024-05-18
Race-condition with fan-in pattern


emil14:
Every time you have `N>1` senders for `1` receiver you can have race-condition

- sender `s1` sent a message `m1` at `0:00:00`
- sender `s2` sent a message `m2` at `0:00:01`
- message `m2` received by receiver `r1` before `m1` - _data-race_

The reason for that is the way runtime interprets the program

For each connection there is a separate goroutine. Goroutines are _concurrent_ which means if we have 2 goroutines they _compete_ for resources.

So, 2 goroutines await for signal from sender (those are different) to deliver a message to receiver (which is the same). Even though first sender sent first _there's no guarantee Go scheduler will activate corresponding connection goroutine first_. Sometimes it does, sometimes it doesn't.

Here's the example

```neva
import { lists }

const lst list<bool> = [true, false]

component Main(start) (stop) {
	nodes { lists.For{Printer} }
	:start -> ($lst -> for -> :stop)
}

component Printer(data bool) (sig any) {
	nodes { If, Println }
	:data -> if
	if:then -> ('true' -> println)
	if:else -> ('false' -> println)
	println -> :sig
}
```

It's clear that `true` must be printed first. Yet it's not always the case.

This behaviour considered a bug that is caused by incorrect implementation. When there will be a language spec we will describe correct behaviour and each implementor will follow it.

We need to **serialize** message passing for connections that follow **fan-in** pattern.

---

**This is root-cause** of #575 and #635

emil14:
# Shared Queue As a Serialization Mechanism

Basically each solution we can get will be serialization for connections with shared receivers. To do so, we need to get rid of concurrency for them.

The most simple way to do is this:

1. Get rid of goroutine-per connection
2. Implement queue and share it across _all_ senders
3. When sender sends to a queue, there is a (one) goroutine that reads from it and does broadcasting

```go
connect() {
	for msg := range q {
		broadcast(msg)
	}
}
``` 

However, there's a couple of problems with this solution...

## Problems

1. It's a bottleneck, each sender sends to the same queue. Slow receiver will slow down fast senders and other receivers.
2. Much bigger problem (not performance, but _correctness_) - how to handle intermediate connections?

### Performance Problem (Making the bottleneck wider)

There's basically 3 ways to improve performance of this solution

1. Add buffer to queue
2. Add buffers to receiver channels
3. Use several queues instead of one

[...]

PR #663, opened 2024-05-28
Race condition fix. Part 1 (Compiler)

emil14:
Implementation of https://github.com/nevalang/neva/issues/644#issuecomment-2138188931


emil14:
Now we need to rewrite runtime so all senders share the same receiver

emil14:
I will not wait for CI cuz race is still there. We need to update runtime

PR #670, opened 2024-06-01
Race condition fix. Part 2 (Runtime)

emil14:


emil14:
TODO

- support event listener
- disable graph reduction with `-debug` flag

PR #679, opened 2024-06-11
Add `{lists,strings}.At` flow ops to access elements of a List and runes in a string

Catya3:
Removes built-in Index

Needed for https://github.com/Catya3/mandelbrot


Catya3:
I realized we have `lists.Index` already

Catya3:
Uhh we still need this. I don't believe `Index` works for me as the generic types are not exact.
```
// Index returns the element at the given index in the ordered collection.
// If the index is out of bounds, it returns an error.
// The index is zero-based.
#extern(index)
pub flow Index<T list<any> | string>(data T, idx int) (res T, err error)
```

res T is a list<any> not an any or even $element_type

Issue #689, opened 2024-07-02
Out of Order Delivery


emil14:
If all you want is to understand the issue, **you can skip the Background** (it's an interesting concurrency journey though!) and start from the description of the problem.

# Background

It all started with #575 - simple task like "print each item of a stream and then terminate" turns out to be not that simple, sometimes program was terminating too soon.

Our first thought was that we need higher order components that can do proper iteration and terminate predictably by utilizing locks internally. Long story short - it turns out to be a rabbit hole.

We added new shiny components like `For` and `Map` and even starting to dream about `Filter` and company. Simple examples was working fine and everything seemed great until @dorian3343 showed couple of little more complex examples where problem arrive again. Honestly at that point I was frustrated. This is what #635 was about.

After several pair programming sessions with @Catya3 we found out that problem is not about For or Map or any other component. It's deeper. It's about runtime.

## Fan-In Problem

This is how #644 started. TLDR: runtime operated in terms of outgoing connections. It means that at runtime connection was `sender -> receiver[]`. In other words, network was a _fan_out_ map. And for each such connection we have a goroutine.

In Neva network graph is a many-to-many relation. Not just one sender can have many receivers but also one receiver can have many senders. It means that not just fan-out (split) but also fan-in (join) is possible. And that's a big problem was this design. It means we can have 2 concurrent goroutines sharing the same receiver.

The simplest case is 2 one-to-one connections like this:

```
s1 -> r1
s2 -> r1
```

`s1` might send to `r1` _before_ `s2` but (if sending happens _fast enough_) there might be _race_ (because goroutines are concurrent). Go scheduler might activate `s2->r1` goroutine before `s1->r1` and `r1` will receive messages _out of order_. By "out of order" I mean order that is different that order of sending.

### First Attempt To Solve (Single Queue)

Goroutines was source of race condition. We need to get rid of them and that will fix out of order right? Well yeah but...

After short thinking decision was made - let all outports send to single queue. If s1 sends to queue it blocks s2, s2 awaits for s1 to send. Then there's a one single goroutine that reads messages that from queue and delivers them to corresponding receivers. This is a relatively simple design but it has 2 problems:

**Problem 1. Intermediate Connections**

Neva program usually include not just "atomic" components (runtime functions) but also "complementary" ones - components implemented in Nevalang itself. They are only ports and connections at runtime. No runtime function is spawned to represent them. They "do nothing" except message passing.

Let's say we have this program:

```neva
flow Main(start) (stop) {
[...]

PR #696, opened 2024-09-15
Connectionless

### LINKED ISSUE #695: Race-free Parallel Connections
Current runtime implementation would spawn goroutine for each pipe/fan-in connection. Even though (direct) fan-in connections are race-free, there's still race-condition in parallel connections.

Simplest picture of the problem. `s1`, `s2`, `r1` and `r2` are all ports:

```neva
s1 -> r1
s2 -> r2
```

Because `s1 -> r1` and `s2 -> r2` are (concurrent) goroutines there's a race-condition. What it means is that this situation is not impossible:

0. 2 goroutines were spawned to serve these two pipelines
1. `s1` sends to `m1` message to `r1` and blocks
2. `s1` sends to  `m2` message to `r2` and also blocks
3. go scheduler activates second `s2 -> r2` goroutine and receives message from `s2` to send it to `r2` (and blocks)
4. `s2` unblocks
5. go scheduler activates first `s1 -> r1` goroutine, and receives message from `s1` to send it to `r1` (and blocks)
6. `s1` unblocks

At the step 3 we can see race condition. There's some order set by senders: `1) s1-> 2) s2->` and that order needs to be preserved in `1) ->r1 2) ->r2`.

It's very important to point that even if this is preserved, there's no guarantee that `r1` will actually receive before `r2`. There are multiple reasons for that starting from `r1` could simply be slower than `r2`, to go scheduler activating r1 and r2 in different order. That's ok, we don't have control over these things.

**What we want guarantee is, again, that we send to receivers in the same order as we received from senders.**

Now it's a lot of questions like:

1. Is this always necessary? Could it be limited to only ports of the same node? E.g. what is the case for `a -> n1:x; b -> n2:x;` where it's critical that `n1` receives before `n2`.
2. If it's necessary (there are cases where >=2 nodes needs order), could it be handled thought locks, HOCs?
3. For what's necessary - which options do we have that meet these requirements? What are their upsides and downsides?

---

Very close to #689

emil14:
- Fixes #695 and #689
- New runtime design
  - No connections, only function calls
  - No native fan-in or fan-out, special function is used for that now
    - Fan-in and Fan-out are implicitly supported by desugarer in compiler
  - Graph reduction is enabled by default and cannot be disabled, program contains no intermediate connections
    - Since any connection is a pipeline between two runtime functions, **each channel represents sender-receiver pair** (branch is called _connectionless_, but connections are actually represented by go channels, there's just no special abstraction for that anymore)
  - Massive refactor everywhere (except stdlib), especially in runtime's function IO API


emil14:
> each channel represents sender-receiver pair 

basically this is the fix. ofc for doing that we had to get rid of fan-in/out so we moved that part from runtime to desugarer (and ofc added respectful components)

PR #745, opened 2024-11-02
feat(std): rename operators/reducers ports nameing from acc/el to lef…

emil14:
…t/right


emil14:
This PR adds `switch` statement and lots of other fixes and improvements

Issue #802, opened 2024-12-13
Routers and Selectors. Part 1 `if`


emil14:
This is a first part in the series of drafts that explores the syntax-level features needed to handle dataflow use cases, beyond the already supported capabilities like fan-in/out, binary/ternary operations, ranges, and struct selectors.

We focus on two main categories of features:

1. Routers - Components that choose a direction for message flow
2. Selectors - Components that choose a specific value

Routers:

- If - Basic conditional routing
- Switch - Multi-branch routing
- Race - Order-based routing

Selectors:

- Match - Value-based selection
- Select - Order-based selection

These components can be further categorized by how they make decisions:

1. Value-based - Uses the content of messages to make choices (If, Switch, Match)
2. Order-based - Uses message arrival order/source to make choices (Race, Select)

An important design principle is that all these features must be race-condition free and intuitive to use. For example, instead of extended switch form, deferred connections could theoretically be used, but they introduce intermediate lock nodes with latency that can cause race conditions. By providing first-class routing constructs instead, we ensure safe and predictable behavior without needing complex locking mechanisms.

### Basic

If is the simplest router that receives a boolean message and routes it into two different directions based on its value.

```
sender -> if {
    then -> then_receiver
    else -> else_receiver
}
```

Example

```
true -> if {
    then -> println
    else -> panic
}
```

### One Branch, Many Receivers

If supports multiple receivers per branch, which are handled like a fan-out. Keep in mind that the If component will not receive new messages until the previous ones are handled. Therefore, the sender of the If will be blocked by the slowest receiver in a branch.

```
sender -> if {
    then -> [then_receiver1, then_receiver2]
    ...
}
```

Example

```
job:failed -> if {
    then -> log:fatal
    else -> [log:info, db:write]
}
```

### With Final Receiver

The If component has an extended form where its body is connected to a receiver. In this case, If sends a message to the selected branch receiver first, and then, after that receiver has received the message, it sends to the final receiver.

It is important to note that the basic If, after receiving a boolean message, is blocked only by the selected branch receiver(s). However, If with a final receiver is also blocked by that final receiver. Similarly, final receiver is blocked by the selected branch receiver.

This setup allows answering the question "when has the selected branch received the message?" but not "when has the selected node finished processing?" For example, sending a message to `println` will start printing, but in `if {...} -> final_receiver`, the final receiver has no guarantee that printing is finished.

```
sender -> if {
    then -> then_receiver
    else -> else_receiver
} -> final_receiver
```

[...]

PR #803, opened 2024-12-13
Routers and Selectors Draft

emil14:


emil14:
Let's this discussion be just an issue https://github.com/nevalang/neva/issues/802

PR #836, opened 2025-01-21
fix: File lock not released properly

Mr-Ao-Dragon:
The original lock management process would first try to `remove` the lock file and then `close` it. This led to the lock file not being removed properly and resulted in a **panic**. I reversed the order, so now it works normally.

### DIFF
--- a/internal/builder/lock.go
+++ b/internal/builder/lock.go
@@ -23,10 +23,12 @@ func acquireLockFile() (release func(), err error) {
 		)
 		if err == nil {
 			return func() {
-				if err := os.Remove(filename); err != nil {
+				//must close first, then remove it.
+				if err := f.Close(); err != nil {
 					panic(err)
 				}
-				if err := f.Close(); err != nil {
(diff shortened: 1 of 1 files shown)


Mr-Ao-Dragon:
Sorry, I forgot to add the GPG digital signature to the submission.

internal/builder/lock.go
```diff
@@ -23,10 +23,12 @@ func acquireLockFile() (release func(), err error) {
...
 		if err == nil {
 			return func() {
-				if err := os.Remove(filename); err != nil {
+				//must close first, then remove it.
+				if err := f.Close(); err != nil {
 					panic(err)
 				}
-				if err := f.Close(); err != nil {
+				if err := os.Remove(filename); err != nil {
+					time.Sleep(0)
```
emil14, line 31: why do we need sleep(0)?

Mr-Ao-Dragon: 我忘删了

PR #841, opened 2025-01-22
add: dynamic template function

Mr-Ao-Dragon:
- Added an advanced template system.
- Update hello-world to the latest version.
- Add an empty program to quickly start writing.
- Added a package template.

### DIFF
--- a/internal/cli/new.go
+++ b/internal/cli/new.go
@@ -2,11 +2,12 @@ package cli
 
 import (
 	"fmt"
+	"github.com/nevalang/neva/internal"
 	"os"
 	"path/filepath"
 
 	"github.com/nevalang/neva/pkg"
-	cli "github.com/urfave/cli/v2"
+	"github.com/urfave/cli/v2"
 )
 
 func newNewCmd(workdir string) *cli.Command {
@@ -19,21 +20,36 @@ func newNewCmd(workdir string) *cli.Command {
 				if err := os.Mkdir(pathArg, 0755); err != nil {
 					return err
 				}
-				return createNevaMod(pathArg)
+				return createNevaMod(pathArg, "hello-world")
 			}
-			return createNevaMod(workdir)
+			return createNevaMod(workdir, "hello-world")
+		},
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:     "template",
+				Usage:    "Specify the template for creating new projects.",
+				Required: false,
+
+				Action: func(context *cli.Context, templateName string) error {
+					if templateName == "" {
+						templateName = "hello-world"
+					}
+					pathArg := context.Args().First()
+					if err := os.Mkdir(pathArg, 0755); err != nil {
+						return err
+					}
+					return createNevaMod(pathArg, templateName)
+
+				},
+			},
 		},
 	}
 }
 
-func createNevaMod(path string) error {
+func createNevaMod(path string, templateName string) error {
 	// Create neva.yml file
 	nevaYmlContent := fmt.Sprintf("neva: %s", pkg.Version)
-	if err := os.WriteFile(
-		filepath.Join(path, "neva.yml"),
-		[]byte(nevaYmlContent),
-		0644,
-	); err != nil {
+	if err := os.WriteFile(filepath.Join(path, "neva.yml"), []byte(nevaYmlContent), 0644); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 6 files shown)


internal/cli/new.go
```diff
@@ -19,21 +20,36 @@ func newNewCmd(workdir string) *cli.Command {
...
+
+				Action: func(context *cli.Context, templateName string) error {
+					if templateName == "" {
+						templateName = "hello-world"
+					}
+					pathArg := context.Args().First()
+					if err := os.Mkdir(pathArg, 0755); err != nil {
+						return err
+					}
+					return createNevaMod(pathArg, templateName)
```
emil14, line 41: Did you test it? Asking because I was thinking `Action` work different. Looks like we call `createNevaMod` two times: from action of the command and form action of the flag. Take a look at how I use flag-actions in other commands (just to set variable and then use it command-action)

internal/templates/empty.neva
```diff
@@ -0,0 +1,11 @@
+// import as here
+// eg. import { fmt }
+
+def Main(start any) (stop any) {
+    // code as here
```
emil14, line 5: `// code as here` -> `// define nodes here`

internal/templates/empty.neva
```diff
@@ -0,0 +1,11 @@
...
+
+    ---
+    // flow as here
```
emil14, line 8: `// flow as here` -> `// define network here`

internal/templates/hello-world.neva
```diff
@@ -0,0 +1,7 @@
+import { fmt }
+
+def Main(start Any)(stop Any) {
+    println fmt.println<string>
```
emil14, line 4: `fmt.println` -> `fmt.Println`

internal/templates/hello-world.neva
```diff
@@ -0,0 +1,7 @@
...
+    ---
+    :start -> "hello world!" -> println -> stop
```
emil14, line 6: `stop` -> `:stop`

internal/templates/hello-world.neva
```diff
@@ -0,0 +1,7 @@
+import { fmt }
```
emil14, line 1: Please move to `internal/cli/templates`, this and the next one

internal/templates/hello-world.neva
```diff
@@ -0,0 +1,7 @@
...
+
+def Main(start Any)(stop Any) {
```
emil14, line 3: `(start Any)(stop Any)` -> `(start any) (stop any)`

internal/templates/package.neva
```diff
@@ -0,0 +1,7 @@
+// don't forget modify i/o type and name
+
+pub def YourNode(in Any)(out Any) {
```
emil14, line 3: `def YourNode(in Any)(out Any)` -> `def YourComponent(data any) (res any)`

internal/templates/package.neva
```diff
@@ -0,0 +1,7 @@
+// don't forget modify i/o type and name
```
emil14, line 1: 1. looks like this template won't compile because it lacks `Main` component
2. name `package` is a little bit ambiguous. what exactly this template should be?

emil14:
Hey there, thanks for the PR. Please answer these two questions, before we merge this:

1. What problem are we trying to solve?
2. Can we add e2e tests that ensure that this functionality works?

Number `1` is the most important, please do not proceed to make changes before we agreed on that 🙏

internal/templates/package.neva
Mr-Ao-Dragon: This template is intended to be reference ready and is not intended to be compilable

internal/templates/hello-world.neva
Mr-Ao-Dragon: Maybe I should just move all the templates to internal\cli\templates?

Mr-Ao-Dragon:
Let me fix the syntax of the template first

Mr-Ao-Dragon:
> Can we add e2e tests that ensure that this functionality works? 

yes, i will

Mr-Ao-Dragon:
> What problem are we trying to solve?

By adding this feature, users can choose their own faster templates for development, and we don't have to do anything to new.go when we add new templates, we can just write them.

internal/templates/hello-world.neva
emil14: Yes, that's was my intention :)

internal/templates/package.neva
emil14: So what will happen if user will run `neva build --template=package`? It will create neva module without executable package? It's like for developing libraries?

emil14:
> By adding this feature, users can choose their own faster templates for development, and we don't have to do anything to new.go when we add new templates, we can just write them.

Let's discuss on Discord

Mr-Ao-Dragon:
neva doesn't have an ecosystem at the moment and this pr is currently unnecessary, so I decided to close it.

Issue #666, opened 2024-06-01
Iterator pattern (early termination for stream processing for both parent and child)


emil14:
# Go Example

In Go we can `break` the loop to stop next `scanner.Text()` calls.

```Go
func main() {
    stream := strings.NewReader("line1\nline2\nline3\nstop\nline4\n")

    scanner := bufio.NewScanner(stream)

    for scanner.Scan() {
        line := scanner.Text() // this is our `->:next`
        fmt.Println("Received:", line)

        if line == "stop" { // this is our `->:done`
            fmt.Println("Terminating the loop.")
            break
        }
    }
}
```

Can we do something like that in Neva? It made me think about termination of a stream processing. It seems like if we have stream we need take it all. Yes we can ignore some elements e.g. process first 10 and ignore all the others but upstream will still produce those elements. This is the core idea of this - **can we make upstream stop**?

# Idea

```neva
// Iterator allows you do early return while you process stream
// The difference with just ignoring the rest of the stream elements
// Is that you can actually stop the stream from happening
// It's important because it means that upstream component can actually stop producing new items
flow Iterator<T>(chunk T) (next any, done any) {
    // receiver chunk
    // send to next if you want next chunk
    // send to done if you done

    // after you send to next you'll receive either
    // next chunk or stream will be ended
    
    // if you'll send to done stream will be ended
    // if there was elements in the stream they will be omited
}
```

Ofc to make it work upstream component must support this "protocol". This is **two-way communication** and in dataflow we need a _loop_ for that (unlike call-return model which is sync+bidirectional).

<img width="362" alt="Screen Shot 2024-06-01 at 15 11 36" src="https://github.com/nevalang/neva/assets/6453537/9ca198f4-5121-426b-9240-a233398e0738">

emil14:
related to #40 #711 and #723

emil14:
maybe it's possible to introduce some kind of generic `for` loop that will support this iterator protocol? something similar to Go 1.23 iterators

emil14:
might also be related to #687

emil14:
other name for this could be "generator"

Issue #435, opened 2023-12-31
[Analyzer]: Node DI has no analysis | interfaces are not type-safe


emil14:
1. It's possible to pass DI deps to the node that doesn't expect those deps with such names
2. It's possible to not pass DI deps to node that does expect deps
3. It's possible to pass deps with incompatible interface
  - Different ports
  - Same ports but incompatible types

emil14:
> **It's possible to pass deps with incompatible interface**

This one is **incredibly important** note

We must implement interface sub-typing in a structural way the same way Go does (and the same way Nevalang implements data-types subtyping)

What I mean is that if interface `IFoo` has 1 outport `(bar any)` and your `Foo` has `(bar any, baz Whatever)` then your **Foo implements IFoo**

This is very important to make bridges, streamers and other adapters/decorators work throughout the language

PR #860, opened 2025-02-01
Added err outport for Println.

### LINKED ISSUE #859: Add `err` outport for `fmt.Print` and `fmt.Println`
- update runtime-function
- update stdlib println signature
- update all examples and e2e tests that uses `fmt.Println` (use search across `*.neva`)
  - ignore smoke tests for parser, they don't care about semantics
- make sure all tests are green, update tests if needed

MDH0:
This will resolve #859.

Added an `err` outport to Println and adjusted the go implementation and tests for that.

There is still a problem regarding the For component, because that one doesn't support error handling yet. Should we deal with this yet and if yes, how should we deal with this? Because theoretically (unless I am missing something) we would need an implementation with error handling and an implementation without error handling.

Also another todo is changing the default template and docs for the new Println change.

Another question would be, if we should do the same for Print?


emil14:
@MDH0 

> Added an err outport to Println and adjusted the go implementation and tests for that.

Nice job 👍

> ... problem regarding the `For` component, because that one doesn't support error handling yet

Good catch. So this one is a bit tricky:

1. Eventually all stream processors must have `err` outport, including `For`
2. We gonna have `errors.Lift` and `errors.Must`
  - `errors.Lift` convert component that doesn't have `err` outport to the one that does (error is never sent then)
  - `errors.Must` does the opposite - it converts component with `res, err` to just `res` and panics if wrapped component fails

So you have choice here

- implement `errors.Must` and use it with existing `For`
- implement `errors.Lift` and change `For`, then update for usage where needed

BTW there's an [issue](https://github.com/nevalang/neva/issues/666) related to this (but not about exactly this)

> Also another todo is changing the default template and docs for the new `Println` change.

Default template must be changed because otherwise e2e tests will fail

> Another question would be, if we should do the same for Print?

Please check if Go's `fmt.Print` returns error. I think it does. We follow Go in this regard. If we need to do it, I suggest to make it as a separate task after merging this one

emil14:
@MDH0 

I believe `errors` package does not exist in `main`, here's the content for `errors/errors.new` as I see it. In case you'll need it

(I didn't test it!)

```neva
import { runtime }

// Lift converts node that doesn't have error outport to one that does.
// The error outport will forever be blocked.
def Lift<T, Y>(data T) (res Y, err error) {
    handler ILiftHandler<T, Y>
    del Del
    ---
    :data -> [
        handler -> :res,
        false -> switch {
            true -> :err
            _ -> del
        }
    ]
}

pub interface ILiftHandler<T, Y>(T) (res Y, err error)

// Must converts node that have error outport to one that doesn't.
// If handler sends an error, the node will panic.
def Must<T, Y>(data T) (res Y) {
    handler IMustHandler<T, Y>
    panic Panic
    ---
    :data -> handler
    handler:res -> :res
    handler:err -> panic
}

pub interface IMustHandler<T, Y>(T) (res Y, err error)

```

MDH0:
Ok, I have more questions regarding this. So first of all, I just checked the Go documentation for their Print's. They throw errors, but they mention, that it is convention, not to worry about errors. There would be the question, how much do we want to follow Go for that? If we really are strict about it, this PR should be closed and we would need to remove the error from Printf.

But independent of this, we have to think about how do deal with For. Because this raises even more questions. The For-handler can throw multiple errors. Therefore, wouldn't we need an error-stream for that?

Alternative ideas I had would be a separate TryFor-component (although I don't really like this idea) or to include an error handler that handles the error seperately (but is against the idea of handling errors in the main component, but might be useful when implementing logging).

emil14:
> They throw errors, but they mention, that it is convention, not to worry about errors

1. Go allows to omit error handling, we don't
2. Go doesn't have `?` and require `if err != nil` boilerplate

Given these statements I would say that both `Print` and `Printf` errors should have `:err` outport:

- Without `err` there's _no way_ of handling the error, unlike Go where you can omit but if needed can handle
- Single interface for stream-processors: `for/map/filter/reduce/etc` can just take `Println` if needed, without any wrappers
- Cognitive overhead is one `?` per node, I consider this small

> For-handler can throw multiple errors. Therefore, wouldn't we need an error-stream for that?

Not really. As soon as `err` happens, stream-processor must stop processing next items in a stream, until this stream is closed and new stream is opened. That is, each `err` message represents a stream, so we don't need `stream<error>`.

Please note that re-implementing stream processors so they follow this logic is a separate task #861 because it requires internal state

> separate TryFor-component

I would rather try to keep APIs minimal and composable instead. E.g. having only one version of `For` - the one that can handle errors. If our handler doesn't return error (which is weird, because for is for side-effects) we can wrap it `For{errors.Lift{MyNode}}`

> or to include an error handler that handles the error seperately (but is against the idea of handling errors in the main component, but might be useful when implementing logging).

The problem with this solution is that optional dependencies are not supported - if component needs dependency, it always needs it. At least this is the current situation

MDH0:
Alright, so we are on the same page, I also change the For-component, to return an error if the handler returns an error and in a later step, errors.Lift will be implemented, so components without an error outport will still be able to be used in For.

emil14:
> As soon as `err` happens, stream-processor must stop processing next items in a stream, until this stream is closed and new stream is opened

By "stop" I mean omit/discard/delete rather than block

MDH0:
New commit includes the changed For component with error handling.

Open Todos:

- Fix documentation and startup project for new Println implementation
- Add error handling to the Tap component without causing deadlocks

MDH0:
@emil14 Can you check the documentation on the new commit?

internal/cli/new.go
```diff
@@ -55,7 +55,8 @@ func createNevaMod(path string) error {
 def Main(start any) (stop any) {
 	fmt.Println
 	---
-	:start -> 'Hello, World!' -> println -> :stop
+	:start -> 'Hello, World!' -> println
+	[println:res, println:err] -> :stop
```
emil14, line 59: let's panic

docs/tutorial.md
```diff
@@ -107,7 +107,8 @@ import { fmt }
 def Main(start any) (stop any) {
 	println fmt.Println<string>
 	---
-	:start -> 'Hello, World!' -> println -> :stop
+	:start -> 'Hello, World!' -> println
+	[println:res, println:err] -> :stop
```
emil14, line 111: let's panic here

emil14:
> @emil14 Can you check the documentation on the new commit?

@MDH0 sure

optional: I suggest to `panic` on `println:err` but we would have to update `README.md` and `docs` so I can do that myself later

e2e/for_with_range_and_if/e2e_test.go
```diff
@@ -8,6 +8,7 @@ import (
 )
 
 func Test(t *testing.T) {
+	t.Skip("Remove this skip, until errors.Lift has been implemented.")
```
emil14, line 11: we going to cover it in this PR right?

e2e/for_with_range_and_if/main/main.neva
```diff
@@ -3,9 +3,11 @@ import { lists, fmt }
 const lst list<bool> = [true, false]
 
 def Main(start any) (stop any) {
-	ListToStream<bool>, For<bool>{PrintAsNum}, Wait
+	ListToStream<bool>, For<bool>{PrintAsNum}, Wait, Panic
```
emil14, line 6: I've found the reason of deadlock - here `PrintAsNum` has outports `sig, err` while `For` expects `res, err`. So println sends to `:sig` but nobody reads from there. It's specific case of the #435 

Here's the correct implementation:

```neva
import { lists, fmt }

const lst list<bool> = [true, false]

def Main(start any) (stop any) {
	ListToStream<bool>, For<bool>{PrintAsNum}, Wait, Panic
	---
	:start -> $lst -> listToStream -> for
	for:res -> wait -> :stop
	for:err -> panic
}

def PrintAsNum(data bool) (res any, err error) {
	println fmt.Println?
	---
	(:data ? 1 : 0) -> println -> :res
}
```

Also no need for `t.Skip()` for this test

emil14:
@MDH0 unfortunately you have to merge `main` branch and resolve conflicts in 3 files after I added `errors` package. I think conflicts should be small, I tried to touch as small stuff as possible

emil14:
I was able to catch deadlock with `neva run -trace e2e/interface_with_imports/main`, trace looks like this:

```log
sent | :start | {}
recv | sub_component/printer/println:data | {}
sent | sub_component/printer/println:res | {}
recv | sub_component/printer/__del__:data | {}
```

MDH0:
Why are the examples compiling now? This is the opposite of 'it works on my system'.

docs/book/networks.md
```diff
@@ -426,17 +427,17 @@ Here's an example using this feature:
 
 ```neva
 def Foo(data) (sig) {
-    Println
+    Print
```
emil14, line 430: why change?

PR #943, opened 2025-11-23
Gemini desugared visualizer experiment

emil14:


internal/compiler/backend/threejs/backend.go
```diff
@@ -0,0 +1,740 @@
...
+)
+
+type Backend struct{}
+
+func NewBackend() Backend {
+	return Backend{}
+}
+
+func (b Backend) Emit(dst string, prog *ir.Program, trace bool) error {
+	return nil
```
emil14: code smell

internal/compiler/backend/threejs/backend.go
```diff
@@ -0,0 +1,740 @@
...
+		// Determine the node name for this function call
+		name := ""
+		if len(f.IO.In) > 0 {
+			name = getNodeName(f.IO.In[0].Path)
+		} else if len(f.IO.Out) > 0 {
+			name = getNodeName(f.IO.Out[0].Path)
+		}
+		
+		if name == "" {
+			continue
```
emil14: this is weird, we need to return err perhaps

PR #956, opened 2025-12-10
Stdlib fs fix

emil14:


internal/builder/builder.go
```diff
@@ -192,9 +193,10 @@ func New(parser ManifestParser) (Builder, error) {
 		return Builder{}, err
 	}
 
-	stdlibPath, err := rewriteStdlibOntoDisk()
```
emil14: rewriteStdlibOntoDisk perhaps needs to be removed

pkg/os/checksum.go
```diff
@@ -0,0 +1,53 @@
...
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"io"
+	"io/fs"
+	"sort"
+)
+
+// ComputeChecksum computes a checksum of all files in the embedded FS
+func ComputeChecksumForFS(filesys fs.FS) (string, error) {
```
emil14: Add unit test(s) for this function

Issue #996, opened 2026-01-09
Proposal: Optimize Message representation via native Go types


emil14:
## Problem Statement

Currently, the Neva runtime (primarily in `internal/runtime/message.go`) wraps all values in a `Msg` interface and its concrete implementations (e.g., `IntMsg`, `ListMsg`, `DictMsg`). This provides a uniform interface but causes significant memory and performance overhead for data-heavy applications.

A list of integers such as `[1, 2, 3]` is currently represented as a `ListMsg` wrapping a `[]Msg`, where each element is an `IntMsg` wrapping an `int64`. This results in:

1. **Memory Inefficiency**: Each element requires an interface allocation and a wrapper struct, rather than being part of a contiguous native slice.
2. **Access Overhead**: Every operation on an element requires an interface method call or a type assertion, which slows down recursive operations and iteration.

## Proposed Optimization

The goal is to allow composite messages to store native Go types directly when they follow a "primitive-only" or "purely composite" recursive structure.

### Target Representative Structures (Pseudocode)

```
P (Primitive) = int | float | string | bool
S (Slice)     = []P | []C
M (Map)       = map[string, P | C]
C (Composite) = S | M
```

_(Where `int` and `float` map to Go's `int64` and `float64` respectively)._

By allowing the `Msg` container to hold these native Go representations (e.g., `[]int64`, `map[string]string`, or nested native structures), we can significantly "save some time and space" while still satisfying the necessary interface.

### 1. Recursive Operations

This optimization must be carefully designed because the runtime relies on several recursive operations (e.g., `Msg.Equal`, `Match`, `MarshalJSON`). It is not immediately clear how to efficiently support a mix of native Go representations and the existing wrapped representations in these functions without significantly increasing complexity.

### 2. Interface Consistency

The `Msg` interface currently expects methods like `List()` to return `[]Msg` and `Dict()` to return `map[string]Msg`. If we use native representations internally, we need to decide whether to:

- Lazily wrap native types when these methods are called (potentially negating performance gains).
- Modify the interface to return more flexible types, which would have a broad impact on the codebase.

### 3. Broadness of Scope

While the initial focus is on the hierarchical primitives/composites described above, this optimization methodology might eventually be applicable to a broader range of messages (e.g., `StructMsg`, `UnionMsg`).

---

P.S. We could start from supporting basic stuff like `[]bool|int|string` and `map[string, bool|int|string]` and not think about recursion. This already could give some performance gain

PR #1004, opened 2026-01-18
Redesign runtime `Msg`

emil14:
### Motivation
- Replace the previous interface-based message representation with a compact tagged-union to reduce allocations and interface dispatch on hot paths.
- Centralize equality/JSON/match semantics to ensure correctness while enabling faster comparisons and iteration.
- Update runtime transport and native funcs to use the new `Msg` API so hot-path ops (`List()`, `Dict()`, struct/union access) avoid unnecessary boxing.
- Make generated code and compiler templates produce and consume the new representation consistently.

### Description
- Replaced the old `Msg` interface and concrete `*Msg` types with a single `Msg` tagged-union struct and `MsgKind`, plus constructor helpers like `NewBoolMsg`, `NewIntMsg`, `NewFloatMsg`, `NewStringMsg`, `NewListMsg`, `NewDictMsg`, `NewStructMsg`, `NewUnionMsg` and `NewUnionMsgNoData` (see `internal/runtime/message.go`).
- Implemented compact storage for primitives (bits/string) and `val any` for composite values, and centralized `Equal`, `MarshalJSON`, `String`, and `Match` logic to operate on the new `Msg` (see `internal/runtime/message.go`).
- Updated runtime APIs and hot-path code to use the new representation and zero-value `Msg{}` as the sentinel for “no data”, including `program` receive/send paths, `interceptors`, `match`/`switch`/`fan_in` handlers and others (see changes in `internal/runtime/program.go`, `internal/runtime/interceptors.go`, and `internal/runtime/funcs/*`).
- Adjusted compiler backend and templates to emit `runtime.Msg{}` for empty configs, use `NewUnionMsgNoData` for tag-only unions, and to call `.IsStruct()/.Struct()` or `.IsUnion()/.Union()` instead of previous type assertions (see `internal/compiler/backend/golang/backend.go`, `internal/compiler/backend/golang/tpl.go`, and generated exports updates).
- Added unit tests and micro-benchmarks for message operations to validate correctness and measure improvements (`internal/runtime/message_test.go` and `internal/runtime/message_bench_test.go`), and documented the change in `AGENTS.md`.

### Testing
- Ran `make build` (invokes builds/tests); the runtime and most packages compiled and tests ran, but CI-local example `examples/http_get` failed due to external network restrictions (test reports: "Domain forbidden").
- Ran `go test ./...`; the majority of unit and e2e packages passed, but `examples/http_get` failed for the same network restriction, causing an overall test failure.
- Ran `golangci-lint run ./...`; this did not complete due to the linter binary being built with an older Go version than the repository target (Go 1.24 vs 1.25).
- Added benchmarks (`BenchmarkMsgListIter`, `BenchmarkMsgDictLookup`, `BenchmarkMsgEqualList`, `BenchmarkMsgStructGet`) that exercise iteration, lookup, equality, and struct access and are included in the tree for performance evaluation.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_6963a6e35660832d93f6b1ebd45421fb)

===

Below is a detailed explanation of how the current tagged‑union Msg design works (as implemented in the latest commit), with emphasis on the bits/any tricks, negative number handling, and why reflect/unsafe were not used. I’ll also explicitly tie this back to the ideas in the “Proposal: Optimize Message representation via native Go types #996”. All references are to the current repo state you asked about.

1. How the new Msg design works (tagged‑union + bits + any)
1.1. Tagged‑union layout
The new Msg is a single struct that uses:

a tag (MsgKind) to know what kind of value is stored,

a fixed bits uint64 field for fast inline storage of numeric/boolean values,

a str string field for string payloads, and

a val any field for composite payloads ([]Msg, map[string]Msg, StructMsg, UnionMsg).
This is all defined in internal/runtime/message.go.

The key shape is:

type Msg struct {
    kind MsgKind
    bits uint64
    str  string
    val  any
}
Because kind tells you which field is “live,” no interface dispatch is needed to read a value: the accessor (Int(), Float(), List(), etc.) just checks the tag and reads the expected field. The tag check is in mustKind, which panics on invalid access, keeping correctness intact while staying cheap in the hot path.

1.2. The bits trick
The bits uint64 field stores small primitive payloads compactly:

bool is stored as bits = 1 for true, 0 for false.

int64 is stored by casting to uint64 (so it preserves the bit pattern).

float64 is stored using math.Float64bits / Float64frombits (the exact IEEE‑754 bit pattern is round‑tripped).

Constructors show this explicitly:

NewBoolMsg sets bits = 1 for true.

NewIntMsg sets bits = uint64(n).

NewFloatMsg sets bits = math.Float64bits(n).
Accessors reverse this: Bool() checks bits==1, Int() casts bits back to int64, Float() uses math.Float64frombits.

1.3. The any trick
The val any field is used only for composite values:

List() returns msg.val.([]Msg)

Dict() returns msg.val.(map[string]Msg)

Struct() returns msg.val.(StructMsg)

Union() returns msg.val.(UnionMsg)

This keeps composite values “boxed” in a single Msg value rather than through a hierarchy of interface values. The constructors NewListMsg, NewDictMsg, NewStructMsg, NewUnionMsg store the composite payload in val and set the appropriate MsgKind.

1.4. Centralized equality / JSON / match semantics
Equality and JSON formatting are now centralized in Msg, so the runtime doesn’t bounce through type-specific methods and interface dispatch:

Msg.Equal compares tags and then compares the correct backing storage (bits, str, slices/maps, or struct/union types).

Msg.MarshalJSON delegates to the correct representation (primitives, list, dict, struct, union).

Match handles union match logic using tag + optional data check. The logic is equivalent to the previous behavior but performed on the new union value directly (with the hasData bit).

This centralization also makes it easier to reason about correctness: only one place to adjust equality or JSON formatting.

2. How negative numbers are stored
Negative integers are stored through the uint64 cast:

func NewIntMsg(n int64) Msg {
    return Msg{kind: MsgKindInt, bits: uint64(n)}
}
func (msg Msg) Int() int64 {
    return int64(msg.bits)
}
Go’s integer casts preserve the two’s‑complement bit pattern, so a negative int64 value becomes a uint64 with the same raw bits. When it’s cast back to int64, the original negative number is recovered. This is safe and lossless for all int64 values. The behavior is in NewIntMsg and Int() respectively.

3. Why no reflect, no unsafe, no runtime package tricks?
3.1. No reflect
Using reflect would:

reintroduce dynamic dispatch and allocations (especially for composite iteration and equality),

complicate the hot path with type switches on reflect.Kind, and

make correctness harder to guarantee (e.g., reflect.DeepEqual is slower and has edge cases).

The current design uses explicit tags and typed accessors, which is faster and more predictable. All runtime dispatch is already O(1) through a switch on MsgKind. That’s directly visible in Msg.Equal, Msg.MarshalJSON, and Msg.String.

3.2. No unsafe
unsafe could theoretically pack more data or avoid bounds checks, but it risks undefined behavior across Go versions and architectures. Given Neva’s emphasis on correctness and portability (multi‑arch builds in make build), staying in safe Go is consistent with the runtime’s long‑term stability. The new approach already avoids interface allocations without unsafe tricks.

3.3. “runtime” package tricks
Direct runtime package usage (e.g., runtime.memmove or internal layout assumptions) would introduce:

implementation‑detail coupling to Go’s runtime,

stricter version constraints (breaking with Go upgrades),

unclear portability for different architectures.

Given the performance gains already captured by eliminating interface dispatch in message handling, it’s not necessary at this stage.

3.4. Alternatives that were effectively chosen
The tagged‑union approach is essentially the best low‑level alternative to reflect/unsafe in Go:

explicit tagging,

fixed layout for primitives,

minimal boxing for composites.

This is already implemented in Msg and its accessors. It’s a classic approach in high‑performance Go code when you want predictable hot‑path behavior. 

4. Relationship to “Proposal: Optimize Message representation via native Go types #996”
The issue proposes native composite storage (e.g., []int64, map[string]string, etc.) to avoid []Msg and map[string]Msg overhead. The new tagged‑union design does not yet store native composite slices or maps, but it lays the groundwork for that direction by:

Centralizing composite decoding in Msg.List() / Msg.Dict(), so we can later extend them to detect “native list/map” encodings without changing call sites. The accessors are already centralized and are the sole place to retrieve composites.

Making Msg a flat struct, so future composite variants can be stored in val any with new internal types (e.g., []int64 or map[string]string) and recognized in List() / Dict() or in new helper methods.

Centralizing equality and JSON so that future composite variants can be handled without refactoring all runtime funcs or compiler code. The switch on MsgKind is the single dispatch point to extend. 

In other words: the new design doesn’t yet implement the “native composite storage” proposed in #996, but it makes it feasible without another large refactor. The hot‑path methods (List, Dict, Equal, MarshalJSON, Match) are centralized and ready for extension.

What the proposal would still require
To fully implement #996’s idea, you’d need:

additional internal composite variants (e.g., listInt64, listString types),

expanded List()/Dict() to either return []Msg/map[string]Msg or provide “fast path” iterators for native composites,

or introduce new accessors (e.g., ListInt64()), which would be a more invasive API change.

The current design doesn’t do this yet, but it is compatible with the incremental approach suggested in the proposal (e.g., start with []bool|int|string and map[string]bool|int|string and treat them as special composite variants stored in val). The single tagged‑union struct makes that evolution straightforward.

5. Additional design details you asked about
5.1. MsgKind and sentinel values
MsgKindInvalid (tag = 0) is the sentinel “empty” message. Msg.IsValid() simply checks that kind != MsgKindInvalid. This is used in generated code to detect empty results without relying on nil (since Msg is no longer an interface).

5.2. Union “no data” semantics
Unions now track whether they contain data using the hasData bool in UnionMsg. The constructors NewUnionMsg and NewUnionMsgNoData set hasData appropriately. Match logic uses hasData to preserve the pattern‑matching behavior (tag only vs. tag+payload).

5.3. Compiler and generated code changes
The compiler backend now emits:

runtime.Msg{} as the “empty config” sentinel,

runtime.NewUnionMsgNoData for tag-only unions,

typed conversions for list/dict/struct/union fields in getMsgFromGo and getGoFromMsg.
These changes ensure that runtime and generated code align with the new Msg struct representation. 

6. Summary of key benefits (performance rationale)
Eliminated interface dispatch in hot paths like List(), Dict(), Equal(), MarshalJSON().

Reduced allocations: primitives are stored inline in a single struct; composite containers do not require extra wrapper structs per element.

Tighter cache behavior: Msg is a compact struct with small fixed‑size fields and a single any for composites.

These benefits are particularly important in runtime funcs like list and dict operations (e.g., list iteration, dict lookup). The new benchmarks were specifically added to measure this (BenchmarkMsgListIter, BenchmarkMsgDictLookup, etc.).

===

Below is a detailed, source‑anchored explanation of why bits uint64 is used for booleans, what alternatives exist, and why strings remain string rather than []byte in the current design. I’m not modifying any code per your instruction.

1. Why booleans use a 64‑bit field (vs 8‑bit or bit‑packing)
1.1. The actual layout goal: fast, fixed‑shape Msg
The current Msg struct is designed to be a fixed‑shape tagged union with minimal branching, so all payloads (bool/int/float) fit in the same inline slot (bits uint64). This eliminates per‑type allocations and avoids interface dispatch in hot paths. The struct uses one tag plus fixed payload fields: bits uint64, str string, and val any.

1.2. Why bits is always 64‑bit
Using a single 64‑bit slot for all numerics and bool yields:

No extra switching or per‑type storage fields inside Msg (we don’t need bool boolVal, int64 intVal, float64 floatVal, etc.).

Predictable layout across platforms and versions.

Fast accessors: Bool(), Int(), and Float() just reinterpret bits based on the MsgKind.
This behavior is encoded in constructors and accessors:
NewBoolMsg sets bits = 1 if true, 0 otherwise; Bool() reads bits == 1; NewIntMsg casts to uint64, Int() casts back; NewFloatMsg uses Float64bits/Float64frombits.

1.3. “But bool only needs 1 byte”
That is true at the value level, but struct layout and access cost matter more here than absolute byte size:

The struct already contains str string (two machine words) and val any (an interface header). The incremental memory difference between a 1‑byte bool and an 8‑byte uint64 is negligible in the overall Msg footprint.

If we used a smaller bool field and a separate field for int/float, we’d either add more branching or grow the struct with multiple typed fields, which can increase size and reduce cache locality.

A single uint64 keeps the code path uniform and avoids extra branching in hot paths like Equal, String, or MarshalJSON which all use the bits slot directly based on MsgKind.

1.4. Could we bit‑pack booleans more tightly?
Technically yes, but the cost would be:

introducing bit masking/shift ops and/or

extra fields or special‑case code paths
That would add complexity and risk branch mispredicts in the hot path. The current design optimizes for predictability and minimal dispatch, not maximum byte‑level compression.

2. Why string is stored as string and not []byte
2.1. Strings already have efficient representation in Go
A string in Go is a read‑only header (pointer + length) and is immutable. That immutability allows:

safe sharing without copies,

efficient hashing and equality, and

direct JSON encoding without conversions.

The Msg type uses the str string field for string values, and MarshalJSON directly calls json.Marshal(msg.Str()) — no extra conversions needed.

2.2. Using []byte would increase complexity
If strings were stored as []byte, you’d need to:

clone or freeze slices to preserve immutability (to avoid aliasing bugs),

convert []byte to string for JSON or for any string comparison, which allocates unless you use unsafe,

and ensure that code calling Str() doesn’t accidentally mutate internal buffers.

Since many runtime funcs and interceptors treat strings as immutable values (e.g., formatting in interceptors, string concatenations, comparisons), string is the safer, zero‑copy type for “message” semantics. The debug interceptor, for example, formats string messages directly using msg.Str() with quoting; that would require additional conversions if []byte were used internally.

2.3. Would []byte help performance?
Usually, not in this context. The runtime message layer is not doing large byte‑level operations on strings; it primarily passes them through. The more common operations in the runtime are:

equality (Msg.Equal),

JSON serialization,

matching and routing.
These are better served by string. If we introduced []byte, we’d likely pay more conversion costs than we’d save.

3. Alternatives considered and tradeoffs (summary)
Option	Benefits	Downsides
Single uint64 bits for bool/int/float	Fast, uniform, minimal branching	“Wastes” bytes for bool, but overall Msg size dominated by other fields
Separate bool field	Saves bytes for bool	Adds fields/branches, expands struct or complicates accessors
Bit‑packing multiple booleans	Best memory density	Adds complexity and potentially hurts hot‑path CPU performance
string storage	Immutable, zero‑copy, cheap equality	None significant for this use case
[]byte storage	Potentially mutable	Requires conversions or unsafe; higher complexity/risk
4. Direct answer to your concern (in one paragraph)
Yes, booleans are stored in a 64‑bit slot even though they need fewer bits, but the design optimizes for uniform access and minimal branching, and the incremental memory is negligible compared to the string and any fields already present in Msg. The choice avoids multiple payload fields or bit‑packing complexity and keeps Equal/MarshalJSON/String fast and predictable. Similarly, strings are stored as string because it is immutable, efficient for equality/JSON, and avoids conversions; moving to []byte would likely increase allocations or require unsafe without meaningful performance wins in the runtime’s current workloads.


internal/runtime/funcs/match.go
```diff
@@ -75,8 +75,13 @@ func (matchSelector) Create(io runtime.IO, _ runtime.Msg) (func(ctx context.Cont
...
 			}
 
-			if u, ok := resMsg.(runtime.UnionMsg); ok {
-				resMsg = u.Data()
+			if resMsg.IsUnion() {
+				unionMsg := resMsg.Union()
+				if unionMsg.HasData() {
+					resMsg = unionMsg.Data()
+				} else {
+					resMsg = runtime.Msg{}
```
emil14: ???

internal/runtime/funcs/fan_in.go
```diff
@@ -26,7 +26,7 @@ func (fanIn) Create(io runtime.IO, _ runtime.Msg) (func(ctx context.Context), er
 				return
 			}
 
-			if !resOut.Send(ctx, dataMsg) {
+			if !resOut.Send(ctx, dataMsg.Msg) {
```
emil14: ?

internal/runtime/message.go
```diff
@@ -18,223 +19,255 @@ func (o OrderedMsg) String() string {
...
+func (msg Msg) IsUnion() bool {
+	return msg.kind == MsgKindUnion
 }
 
-// --- STRING ---
-type StringMsg struct {
-	internalMsg
-	v string
+func (msg Msg) Bool() bool {
+	msg.mustKind(MsgKindBool, "Bool")
```
emil14: do we need this? I guess this is replacement for old panic from base class

internal/runtime/message.go
```diff
@@ -18,223 +19,255 @@ func (o OrderedMsg) String() string {
...
+	msg.mustKind(MsgKindString, "Str")
+	return msg.str
 }
 
-func (msg StringMsg) Equal(other Msg) bool {
-	otherString, ok := other.(StringMsg)
-	return ok && msg.v == otherString.v
+func (msg Msg) List() []Msg {
+	msg.mustKind(MsgKindList, "List")
+	return msg.val.([]Msg)
```
emil14: I guess we still don't utilize types like `[]int` or `[]string`

internal/runtime/message.go
```diff
@@ -18,223 +19,255 @@ func (o OrderedMsg) String() string {
...
+		return msg.Union().String()
+	default:
+		return "<unknown>"
+	}
+}
+
+func (msg Msg) MarshalJSON() ([]byte, error) {
+	switch msg.kind {
+	case MsgKindInvalid:
+		return []byte("null"), nil
```
emil14: invalid value is not null, even panic is better

internal/runtime/message.go
```diff
@@ -350,70 +382,140 @@ func (msg UnionMsg) Tag() string {
...
 
-func NewUnionMsg(tag string, data Msg) UnionMsg {
+func newUnionMsg(tag string, data Msg) UnionMsg {
 	return UnionMsg{
-		internalMsg: internalMsg{},
-		tag:         tag,
-		data:        data,
+		tag:     tag,
+		data:    data,
+		hasData: data.IsValid(),
```
emil14: oh I see, so "invalid" message is actually usefull... this is not good, we can't panic

internal/runtime/message.go
```diff
@@ -350,70 +382,140 @@ func (msg UnionMsg) Tag() string {
...
+	}
+
+	if !msgUnion.hasData {
+		return true
 	}
 
 	return msgUnion.data.Equal(patternUnion.data)
 }
+
+func equalMsgSlices(left []Msg, right []Msg) bool {
```
emil14: slices->lists

internal/runtime/message.go
```diff
@@ -350,70 +382,140 @@ func (msg UnionMsg) Tag() string {
...
+	}
+	for i := range left {
+		if !left[i].Equal(right[i]) {
+			return false
+		}
+	}
+	return true
+}
+
+func equalMsgMaps(left map[string]Msg, right map[string]Msg) bool {
```
emil14: why do we need this?

internal/runtime/message_bench_test.go
```diff
@@ -0,0 +1,84 @@
+package runtime_test
```
emil14: CRITICAL BLOCKER: we must implement some kind of **E2E** benchmark before we can merge this branch. First we need to add that to `main`, then merge here and only then see improvement. These unit benchmarks _could_ be useful but they are not what we _need_

internal/runtime/message.go
```diff
@@ -18,223 +19,255 @@ func (o OrderedMsg) String() string {
...
-	v string
+func (msg Msg) Bool() bool {
+	msg.mustKind(MsgKindBool, "Bool")
+	return msg.bits == 1
 }
 
-func (msg StringMsg) Str() string { return msg.v }
+func (msg Msg) Int() int64 {
+	msg.mustKind(MsgKindInt, "Int")
+	return int64(msg.bits)
```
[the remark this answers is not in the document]
emil14: Negative ints are stored as the raw int64 two's-complement bit pattern inside uint64 and cast back on read. I also added TestIntMsgNegativeRoundTrip to lock this behavior.

internal/runtime/message.go
```diff
@@ -18,223 +19,255 @@ func (o OrderedMsg) String() string {
...
 }
 
-// --- DICT ---
-type DictMsg struct {
-	internalMsg
-	v map[string]Msg
+func (kind MsgKind) String() string {
+	switch kind {
+	case MsgKindInvalid:
+		return "invalid"
```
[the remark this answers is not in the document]
emil14: Fixed. Invalid Msg no longer serializes as null in this path; it now panics.

internal/runtime/message.go
emil14: Addressed in follow-up changes and now outdated; latest fixes are in commit 4b335dc3.

internal/runtime/message.go
```diff
@@ -18,223 +19,255 @@ func (o OrderedMsg) String() string {
...
 }
 
-func NewStringMsg(s string) StringMsg {
-	return StringMsg{
-		internalMsg: internalMsg{},
-		v:           s,
-	}
+func (msg Msg) Dict() map[string]Msg {
+	msg.mustKind(MsgKindDict, "Dict")
+	return msg.val.(map[string]Msg)
```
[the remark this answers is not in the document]
emil14: Addressed in follow-up changes and now outdated; latest fixes are in commit 4b335dc3.

internal/runtime/message.go
emil14: Addressed in follow-up changes and now outdated; latest fixes are in commit 4b335dc3.

internal/runtime/message.go
```diff
@@ -256,25 +289,23 @@ func (msg StructMsg) get(name string) (Msg, bool) {
...
+	return result
+}
 
-	jsonString := string(jsonData)
-	jsonString = strings.ReplaceAll(jsonString, ":", ": ")
-	jsonString = strings.ReplaceAll(jsonString, ",", ", ")
-
-	return []byte(jsonString), nil
+func (msg StructMsg) MarshalJSON() ([]byte, error) {
+	return marshalMapWithSpaces(msg.Map())
```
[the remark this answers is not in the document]
emil14: Addressed in follow-up changes and now outdated; latest fixes are in commit 4b335dc3.

internal/runtime/message.go
```diff
@@ -350,70 +382,140 @@ func (msg UnionMsg) Tag() string {
 	return msg.tag
 }
 
+func (msg UnionMsg) HasData() bool {
+	return msg.hasData
+}
+
 func (msg UnionMsg) Data() Msg {
+	if !msg.hasData {
+		return Msg{}
```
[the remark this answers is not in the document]
emil14: Addressed in follow-up changes and now outdated; latest fixes are in commit 4b335dc3.

internal/runtime/message.go
emil14: Addressed in follow-up changes and now outdated; latest fixes are in commit 4b335dc3.

emil14:
Heads-up: I split out JSON-format correctness fixes into #1030 (safe spacing outside strings + union data JSON quoting/escaping) so this can merge on NAME:
   neva - Dataflow programming language with static types and implicit parallelism

USAGE:
   neva [global options] command [command options] 

COMMANDS:
   version  Get current Nevalang version
   upgrade  Upgrade to newest Nevalang version
   new      Create new neva project
   get      Add dependency to current module
   run      Build and run neva program from source code
   build    Build neva program
   osarch   List supported OS/architecture combinations for native target
   doc      Search the standard library for symbols using regular expressions
   help, h  Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --help, -h  show help without waiting for message-representation redesign.\n\nAfter rebasing #1004, please port/cherry-pick equivalent changes from #1030 to preserve the same formatting guarantees in the new  implementation.

emil14:
Heads-up: I split out JSON-format correctness fixes into #1030 (safe spacing outside strings plus union data JSON quoting and escaping) so this can merge on main without waiting for message-representation redesign.\n\nAfter rebasing #1004, please port or cherry-pick equivalent changes from #1030 to preserve the same formatting guarantees in the redesigned message implementation.

internal/runtime/funcs/http.go
```diff
@@ -66,7 +66,7 @@ func (httpGet) Create(funcIO runtime.IO, _ runtime.Msg) (func(ctx context.Contex
 	}, nil
 }
 
-func respMsg(statusCode int, body []byte) runtime.StructMsg {
+func respMsg(statusCode int, body []byte) runtime.Msg {
 	return runtime.NewStructMsg([]runtime.StructField{
 		runtime.NewStructField("body", runtime.NewStringMsg(string(body))),
```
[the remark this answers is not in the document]
emil14: Good catch. I verified this against current `main`: `std/http.Response.body` is `bytes`, and this path already sends `runtime.NewBytesMsg(body)`, so no additional code change was needed here.

internal/runtime/message.go
```diff
@@ -256,24 +308,53 @@ func (msg StructMsg) get(name string) (Msg, bool) {
...
+		bb.Write(nameJSON)
+		bb.WriteByte(':')
+		bb.Write(valueJSON)
 	}
 
-	jsonString := string(jsonData)
+	bb.WriteByte('}')
+
+	jsonString := bb.String()
 	jsonString = strings.ReplaceAll(jsonString, ":", ": ")
```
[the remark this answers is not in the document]
emil14: Addressed in commits `f9cee287` + `f08d9ef3`. I removed unsafe whole-string post-processing (`addJSONSpaces`) and switched to direct JSON writers for dict/struct output. For nested values, serialization is now context-aware (`marshalNestedJSON`) so string payloads are never rewritten after marshaling, while preserving existing runtime output contracts used by e2e/examples.

emil14:
> Replying to @github-actions opencode review (2026-04-12)
>
> "JSON spacing normalization still unsafe"

Handled: removed `addJSONSpaces` and replaced with direct JSON emission paths in runtime message serialization (`f9cee287`, `f08d9ef3`).

> "Missing E2E benchmark"

Confirmed already in place earlier in this PR track; local validation also included benchmark execution (`go test -run ^ -bench ...`).

> "HTTP body handling"

Verified current behavior matches `main`: `std/http.Response.body` is `bytes` and runtime path sends `NewBytesMsg(body)`.

emil14:
Finalization update for #1004 (head `670fbfe`).

### What was done
- Merged `origin/main` into `codex/redesign-runtime-message-representation` and resolved conflicts in:
  - `internal/runtime/message.go`
  - `internal/runtime/program.go`
  - `internal/runtime/funcs/switch.go`
  - `internal/compiler/utils/generated/exports.go`
- Regenerated compiler glue via repo CLI (not global binary):
  - `go run ./cmd/neva build --target=go --target-go-mode=pkg --target-go-runtime-path=../runtime --output=utils/generated utils` (from `internal/compiler`)
- Kept tagged-union `Msg` design and aligned merge resolution around it.
- Resolved all previously open review threads in this PR.

### Validation
- Local targeted checks:
  - `go test ./internal/runtime/...`
  - `go test ./internal/compiler/...`
  - `golangci-lint` on runtime + go backend paths (clean)
- GitHub checks on current head: all required checks are green (`gofix-check`, `golangci`, `govulncheck`, `unit_tests`, `e2e_tests`).

### Benchmark proof (agreed protocol)
Baseline: `origin/main`

Commands used:
- `go test ./benchmarks -run '^$' -bench '^BenchmarkRuntimeE2E$' -benchmem -count=3`
- `go test ./internal/runtime -run '^$' -bench '^BenchmarkMsg' -benchmem -count=5`
- `benchstat` over raw outputs

Results:
1. **Runtime E2E (`main` vs this branch)**
   - `benchstat` geomean: **-8.33% sec/op** (branch faster on geomean)
   - B/op geomean: **+0.21%** (effectively flat)
   - allocs/op geomean: **no change**
   - Note: with `count=3`, most per-subbenchmark rows are not statistically significant at 95% (benchstat warns about sample size), but there is no signal of broad regressions in allocs/memory and geomean runtime trends positive.

2. **Msg microbench (`BenchmarkMsg*`)**
   - `main` currently has no matching `BenchmarkMsg*`, so cross-branch delta cannot be computed directly for that suite.
   - Current branch absolute results (count=5):
     - `BenchmarkMsgListIter`: ~4.39 µs/op, 0 allocs
     - `BenchmarkMsgDictLookup`: ~17.9 µs/op, 0 allocs
     - `BenchmarkMsgEqualList`: ~2.78 µs/op, 0 allocs
     - `BenchmarkMsgStructGet`: ~74.65 ns/op, 0 allocs

Given merge resolution, green CI, addressed review threads, and benchmark evidence above, this PR is in **ready-to-merge** state.

emil14:
Closing this PR without merge. The implementation is obsolete, but its research context is preserved rather than discarded.

## Why this branch is not a merge candidate

The branch is structurally incompatible with current `main`: it replaces the message model with a flat tagged `Msg`, but its list and dictionary payloads are only `[]Msg` and `map[string]Msg`. That would regress the native scalar storage delivered by #1127 and preserved across compiler and stream materialization by #1157. The branch also predates the accepted `internal/runtime/messages` ownership boundary and changes transport/order/tracing semantics alongside the payload representation.

A rebase would therefore not be conflict resolution; it would be a new architecture project with a high risk of combining two incompatible representations.

## What we retain

- The open hypothesis that a flat tagged scalar value could improve scalar message access, routing, and equality.
- The idea of explicitly measuring accessor and equality hot paths, including cold panic paths.
- Future questions about deeper native composite storage and indexed union tags.
- The JSON lesson from #1030: formatting must be context-aware and must not rewrite serialized string payloads.

The historical May E2E result (`-8.33%` geomean across three runs) is retained only as a directional signal; it is not reproducible merge evidence.

## Where the context now lives

- #996 is the live design record: delivered typed-container work, retained hypotheses, invariants, and explicit non-goals.
- #1015 contains the current experiment protocol, required benchmark matrix, reporting rules, and continuation gate.

Any successor must start from fresh `main`, preserve representation-aware `ListMsg`/`DictMsg`, keep `OrderedMsg`/causes/tracing intact, and prove a reproducible end-to-end gain. It must not rebase or partially port this branch.

Related delivered work: #1127, #1151, #1153, #1154, #1155, #1157.
Related design/performance records: #996, #1015, #1030.

PR #1010, opened 2026-02-01
Fix overload resolution from ambiguous neighbor constraints

emil14:
## Summary
- Fix overload resolution when constraints are derived from neighbors with ambiguous port types.
- Add a new `examples/interest_pipe` example and e2e test demonstrating interest calculation flow.

## Problem
Overload selection for native ops like `Sub`/`Mul` failed with errors like:

```
no compatible overload found for node mul_rate_balance (total components: 2, remaining: 0)
```

This happened when constraints were inferred from neighboring *overloaded* nodes. We collected **all** possible port types and required a candidate to satisfy all of them (e.g., both `int` and `float`), which makes every overload incompatible.

## Fix
When a neighbor node yields multiple possible port types, treat that as **ambiguous** and skip adding those types as constraints. This prevents eliminating all candidates and lets genuine type information (literals, explicit ports, resolved inputs) drive overload selection.

### Code change
`internal/compiler/analyzer/nodes.go`
- `deriveNodeConstraintsFromNetwork`: filter ambiguous neighbor port types before adding outgoing constraints.
- `getPossibleSenderTypes`: filter ambiguous sender port types before adding incoming constraints.

### Before (failure)
```neva
sub Sub
mul_rate_balance Mul
mul_total Mul
---
:state -> [
	.rate -> mul_rate_balance:left,
	.balance -> mul_rate_balance:right,
	.last_time -> sub:right
]
:now -> sub:left
mul_rate_balance -> mul_total:left
sub -> mul_total:right
mul_total -> :interest
```
This could fail if neighbor nodes are overloaded and their port types are ambiguous.

### After (works)
The same code now resolves correctly because ambiguous neighbor types no longer poison overload constraints.

## Example added
`examples/interest_pipe`
- Implements a small interest calculation pipeline with state updates.
- Uses `Sub`/`Mul` directly (overload resolution is exercised).
- Includes `examples/interest_pipe/e2e_test.go` to validate output.

## Tests
- `go test ./examples/interest_pipe`
- `golangci-lint run ./...`

> Full `go test ./...` was started but timed out at the 5-minute cap; it was progressing with no failures at the time.

### DIFF
--- a/internal/compiler/analyzer/nodes.go
+++ b/internal/compiler/analyzer/nodes.go
@@ -4,9 +4,9 @@ import (
 	"fmt"
 
 	"github.com/nevalang/neva/internal/compiler"
-	src "github.com/nevalang/neva/internal/compiler/ast"
-	"github.com/nevalang/neva/internal/compiler/ast/core"
 	"github.com/nevalang/neva/internal/compiler/typesystem"
+	src "github.com/nevalang/neva/pkg/ast"
+	"github.com/nevalang/neva/pkg/core"
 )
 
 type foundInterface struct {
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -137,6 +137,12 @@ Follow these instructions.
 - **Architecture insights**: `deriveNodeConstraintsFromNetwork` needs both direct senders and previous chain-link context to infer selector output types before desugaring.
 - **Gotchas**: Without selector-aware constraints, overloaded std operators can appear ambiguous and force typed wrapper hacks in examples.
 
+- **Common patterns**: New LSP features in `cmd/lsp/server` should include short function doc comments plus targeted inline comments for recursive AST traversal/encoding code.
+- **Common patterns**: For LSP handlers returning `any`, prefer typed empty results (e.g., empty completion/symbol/location lists) or `false` for `PrepareRename` fallback instead of `nil, nil` to satisfy `nilnil`.
+- **Common patterns**: For LSP file lookup failures (`findFile`), propagate the error rather than returning success with nil payload to avoid `nilerr`.
+- **Common patterns**: `gosec` G115 in LSP token/range encoding should use explicit int bounds checks plus `// #nosec G115` on checked casts.
+- **Gotchas**: `nilnil` can be intentionally valid for LSP nullable results (e.g., hover/rename); use narrowly scoped `//nolint:nilnil` with a protocol rationale when required.
+
 ### Session Notes (2026-02-14)
 
 - **Common patterns**: In overload-constraint collection, treat empty or ambiguous neighbor candidate type sets as "no constraint" and let other edges disambiguate overloads.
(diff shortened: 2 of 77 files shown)


internal/compiler/analyzer/nodes.go
```diff
@@ -816,6 +816,30 @@ func (a Analyzer) deriveNodeConstraintsFromNetwork(
...
 
+	// Keep only constraints that are unambiguous across overloads.
+	// Example:
+	//   x Foo          // Foo has overloads Foo(int) and Foo(float)
+	//   x:out -> y:in  // y should not force both int and float as constraints
+	// If a neighbor contributes multiple distinct candidate types, we drop it
+	// from overload constraints and let other edges disambiguate.
+	filterAmbiguous := func(types []typesystem.Expr) []typesystem.Expr {
+		if len(types) == 0 {
+			return nil
```
emil14: ?

examples/interest_pipe/e2e_test.go
```diff
@@ -0,0 +1,14 @@
...
+	"strings"
+	"testing"
+
+	"github.com/nevalang/neva/pkg/e2e"
+	"github.com/stretchr/testify/require"
+)
+
+func Test(t *testing.T) {
+	out, _ := e2e.Run(t, []string{"run", "."})
+	require.ElementsMatch(t, []string{"5000", "5000", "6000"}, strings.Fields(out))
```
emil14: explain in the comments why result is this

internal/compiler/analyzer/nodes.go
```diff
@@ -1016,8 +1057,9 @@ func (a Analyzer) flattenReceiversPortAddrs(receivers []src.ConnectionReceiver)
 }
 
 type receiverSenderPair struct {
-	portAddr src.PortAddr
-	senders  []src.ConnectionSender
+	portAddr      src.PortAddr
+	senders       []src.ConnectionSender
+	prevChainLink []src.ConnectionSender
```
emil14: why? add comment

internal/compiler/analyzer/nodes.go
```diff
@@ -1030,45 +1072,64 @@ type receiverSenderPair struct {
...
 func (a Analyzer) getPossibleSenderTypes(
 	scope src.Scope,
 	parentFrame map[string]typesystem.Def,
 	parentIface src.Interface,
 	nodes map[string]src.Node,
 	sender src.ConnectionSender,
+	prevChainLink []src.ConnectionSender,
 	net []src.Connection,
 ) []typesystem.Expr {
 	// FIXME: looks like we ignore errors here (and in some lower-level functions we call)
```
emil14: Carefully analyze the down call tree and try to resolve this `FIXME` comment by making changes. We should never ignore errors except it's explicitly stated why we do that

internal/compiler/analyzer/nodes.go
```diff
@@ -1085,6 +1146,47 @@ func (a Analyzer) getPossibleSenderTypes(
 		}
 	}
 
+	// struct selector sender in a chain (e.g. :state -> .rate -> mul:left)
+	if len(sender.StructSelector) > 0 {
```
emil14: why wasn't it here before? reply to this comment here in gh PR, explain

internal/compiler/analyzer/nodes.go
```diff
@@ -1030,45 +1100,29 @@ type receiverSenderPair struct {
...
-	}
-	visit(receivers, senders)
+	collectReceiverSenderPairsRec(&pairs, receivers, senders, nil)
 	return pairs
 }
 
 // getPossibleSenderTypes is needed to derive node constraints from the network.
 // It's part of the overloading implementation.
 // It returns a set of possible types produced by a given sender without requiring resolved node interfaces.
+// This function is intentionally best-effort: unresolved pieces return no constraint
```
emil14: how do you know its best effort? its clear for you from the code? explain

internal/compiler/analyzer/nodes.go
```diff
@@ -1121,14 +1209,26 @@ func (a Analyzer) getPossibleSenderTypes(
...
+		if len(types) == 0 {
+			return nil
+		}
+		// Same rule as in deriveNodeConstraintsFromNetwork: ambiguous neighbor output
+		// should not be treated as a hard constraint.
+		// Example:
+		//   maybe_num SomeNode // SomeNode:out can be int OR float due to overloads
+		//   maybe_num -> add:left
+		// We only return a type here when every overload agrees on a single type.
+		return keepSingleUnambiguousType(types)
```
emil14: why don't you think that it must always be clear that its possile to resolve specific overload? is this not how language works? it doesn feel right?

internal/compiler/analyzer/nodes.go
emil14: Clarified in docs. It is best-effort because this function is only used during overload-candidate filtering (`deriveNodeConstraintsFromNetwork`) before full node interfaces are finalized. At this stage we may legitimately fail to derive some sender types, so we return "no constraint" and let the normal validation pass emit concrete diagnostics.

PR #1027, opened 2026-02-18
indexer: keep workspace build when main package is missing

emil14:
## Summary
- fix `pkg/indexer.FullScan` so `main package not found` is treated as a workspace/LSP fallback case
- retry analyzer in library mode (`mainPkgName == ""`) and return that build when fallback succeeds
- keep all other parser/analyzer/frontend failures fatal (no masking)
- add explicit indexer tests for fallback, normal main-package workspace, analyzer-fatal, and parser-fatal scenarios
- add AGENTS session note about the new fallback rule

## Root cause
`FullScan` always used executable analysis with `feResult.MainPkg`. When workspace root was not runnable (for example `examples/`), analyzer returned `main package not found`, and indexer returned an empty build. LSP then failed language features with `file not found in build`.

## Verification
- `golangci-lint run ./...`
- `go test ./...`

### DIFF
--- a/pkg/indexer/indexer.go
+++ b/pkg/indexer/indexer.go
@@ -41,6 +41,10 @@ func (i Indexer) FullScan(
 
 	i.logger.Debug("nevalang module found in workspace", "path", feResult.Path)
 
+	// Frontend derives feResult.MainPkg from the input path (often "." for module
+	// root scans). For workspace indexing we intentionally analyze as library
+	// (empty main package) so LSP build snapshots are not coupled to executable
+	// Main-package validation.
 	aBuild, err := i.analyzer.Analyze(feResult.ParsedBuild, "")
 	if err != nil {
 		return src.Build{}, true, wrapCompilerError(err)
--- a/pkg/indexer/indexer_test.go
+++ b/pkg/indexer/indexer_test.go
@@ -2,20 +2,26 @@ package indexer
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 	"testing"
 
+	neva "github.com/nevalang/neva/pkg"
 	"github.com/stretchr/testify/require"
 	"github.com/tliron/commonlog"
 )
 
+// TestFullScan_workspace_without_main_package_uses_workspace_mode_analysis ensures
+// module-root scans work when Neva code lives in subpackages, e.g.:
+//
+//	neva.yml
+//	pkg/main.neva
 func TestFullScan_workspace_without_main_package_uses_workspace_mode_analysis(t *testing.T) {
 	t.Parallel()
 
 	workspace := writeWorkspace(t, map[string]string{
-		"neva.yml": `neva: 0.34.0
-`,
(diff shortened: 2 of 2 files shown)


pkg/indexer/indexer.go
```diff
@@ -43,12 +45,47 @@ func (i Indexer) FullScan(
 
 	aBuild, err := i.analyzer.Analyze(feResult.ParsedBuild, feResult.MainPkg)
```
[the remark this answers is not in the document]
emil14: Agreed. I updated the inline comment and log message to reflect this as an expected module-root scenario (workspace mode retry), not an exceptional one. This keeps the same behavior while making the intent align with the usual module/package layout.

pkg/indexer/indexer.go
```diff
@@ -43,12 +45,47 @@ func (i Indexer) FullScan(
...
+		// is not itself a runnable entry package.
+		i.logger.Info(
+			"main package not found; falling back to workspace library analysis",
+			"mainPkg", feResult.MainPkg,
+			"workspacePath", workspacePath,
+		)
+
+		fallbackBuild, fallbackErr := i.analyzer.Analyze(feResult.ParsedBuild, "")
+		if fallbackErr != nil {
+			i.logger.Warning(
```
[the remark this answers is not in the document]
emil14: Good call. I removed the warning log from the fallback-failure branch, so that path now only returns the wrapped error without additional logging. We still keep debug-level context logs for fallback activation/success.

pkg/indexer/indexer.go
```diff
@@ -41,7 +41,7 @@ func (i Indexer) FullScan(
 
 	i.logger.Debug("nevalang module found in workspace", "path", feResult.Path)
 
-	aBuild, err := i.analyzer.Analyze(feResult.ParsedBuild, feResult.MainPkg)
+	aBuild, err := i.analyzer.Analyze(feResult.ParsedBuild, "")
```
emil14: explain in comments why including what feResult.MainPkg is and why not pass it instead

pkg/indexer/indexer_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"context"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/tliron/commonlog"
+)
+
+func TestFullScan_workspace_without_main_package_uses_workspace_mode_analysis(t *testing.T) {
```
emil14: Add comments to each test with maybe neva examples whee helps

emil14: Added short doc comments for each indexer test and included concrete Neva workspace layout examples where helpful (module root + subpackage layout).

pkg/indexer/indexer.go
emil14: Added an explicit comment above `Analyze(..., "")` explaining what `feResult.MainPkg` represents (derived from input path, often `"."` at module root) and why workspace scans intentionally do not pass it.

Issue #1033, opened 2026-02-22
P0 milestone: Stdlib usability stabilization tracker


emil14:
## Goal
Create a single P0 tracking issue for all stdlib tasks required to make Neva consistently usable before broader adoption.

## Why this tracker exists
These tasks are highly coupled (API shape, naming, stream semantics, sync/flow ergonomics, package boundaries). Progress should be tracked in one place to avoid context loss across separate implementation windows and autonomous-agent runs.

## Baseline decisions already made (current context)
During the latest stdlib cleanup pass (see #1031), we fixed part of the naming/API inconsistencies and synced docs.

- Port naming baseline: `res` is the primary output, `err` is failure output, `data` is input-only generic placeholder (do not use `data` as outport name).
- Stream iteration baseline: keep `streams.ForEach` as the canonical side-effect iterator; duplicate builtin stream-loop API was removed.
- Documentation baseline: `docs/style_guide.md` and `docs/qna.md` were updated and should be treated as current source of truth for naming/idiomatic usage.

These decisions are considered accepted unless explicitly superseded in a dedicated issue/PR.

## Recommended start context for autonomous work
Do not duplicate long context from repository docs. Start by reading:
- `AGENTS.md`
- `CONTRIBUTING.md`
- `ARCHITECTURE.md`
- `docs/style_guide.md`
- `docs/qna.md`

Then inspect relevant code in:
- `std/`
- `internal/runtime/funcs/`
- `examples/` and `e2e/`

## Sub-issues (linked scope)
- [x] #983 (closed; baseline naming cleanup already landed in #1031)
- [ ] #684
- [ ] #657
- [ ] #864
- [ ] #987
- [ ] #986
- [ ] #1014
- [ ] #583
- [ ] #908
- [ ] #861
- [ ] #666
- [ ] #989
- [ ] #990
- [ ] #647
- [ ] #648
- [ ] #650
- [ ] #671
- [ ] #790
- [ ] #586
- [ ] #664

## Exit criteria for this tracker
- stdlib package taxonomy is explicit and documented (core/prelude/domain or approved alternative);
- stream/sync/flow APIs are coherent and documented;
- major naming/API inconsistencies resolved;
- examples/e2e/docs aligned with final API decisions.

emil14:
[...]

PR #1040, opened 2026-02-24
stdlib/io: add explicit Open/Create/Close file-handle API

### LINKED ISSUE #790: IO API - Nevalang vs Go (descriptions and explicit opening/closing)
Example: our `def ReadAll(filename string) (res string, err error)` works only with fs files while Go's accepts abstract `File` interface that you can/need to get from `os.Open`. Go's API is about passing file descriptors around and explicitly closing them after. This is more verbose but also more flexible. Go's `io.Reader` might work with any streams of bytes, not just fs files.

At this point we must figure out how to implement streaming IO properly and these kinds of things are foundation for that.

emil14:
## Summary
- add explicit file-handle API to `std/io`: `File`, `Open`, `Create`, `Close`, `ReadAllFile`, `WriteAllFile`
- keep existing filename-based `ReadAll`/`WriteAll` intact for backward compatibility
- implement runtime externs with a registry-scoped file-handle store in `internal/runtime/funcs`
- add tests for handle store lifecycle and registry wiring
- add `examples/file_handles` e2e example demonstrating explicit open/write/close/read/close flow

## Context
Part of stdlib usability stabilization: #1033.

Fixes #790.

## Validation
- `golangci-lint run ./...` (pass)
- `go test ./internal/runtime/funcs ./examples/file_handles` (pass)
- `go test -timeout 5m ./...` (known unrelated failures/timeouts in long suites, including `e2e/order_dependend_with_arr_inport` and `examples/delayed_echo`)


emil14 (commented):
# Review Summary
## Critical Issues (Blocking)
- `internal/runtime/funcs/file_read_all_handle.go:67` and `internal/runtime/funcs/file_write_all_handle.go:71`: on read/write failure these components send only `err` and do not emit or close the handle. The API/example pattern consumes the handle and then relies on `ReadAllFile:handle` / `WriteAllFile:res` to reach `Close`, so an I/O error leaves a live descriptor in the registry store with no closeable token in the normal graph. Please either close the handle internally on operation failure, or expose/send a closeable handle on the error path and document that lifecycle rule.

## Non-critical Issues
- `internal/runtime/funcs/file_write_all_handle.go:37`: `WriteAllFile` receives `file` before `data` sequentially. Existing two-input runtime funcs use `receive2` to wait on both inputs concurrently, and the repo Go authoring rules explicitly disallow new sequential two-input receives in runtime funcs. This makes the component unnecessarily arrival-order-sensitive and can introduce avoidable backpressure/deadlock risk in dataflow graphs. Please switch this loop to `receive2(ctx, fileIn, dataIn)`, matching `write_all` and the other two-input externs.
- `.github/workflows/lint.yml:53`: `go run golang.org/x/vuln/cmd/govulncheck@latest ./...` makes CI depend on a moving tool version for every PR run. Please pin the govulncheck module version, or use a pinned action/versioned install path, so lint results and supply-chain inputs are reproducible.

emil14:
Addressed the review findings in 2c37231650eb4c2449a1d059d741ee375a9c33c3:

- `ReadAllFile` and `WriteAllFile` now emit the file handle on operation error paths before sending `err`, so callers can still close the handle.
- `WriteAllFile` now receives `file` and `data` concurrently via `receive2`.
- `govulncheck` is pinned to `golang.org/x/vuln@v1.4.0` in CI and Makefile.

examples/file_handles/main.neva
```diff
@@ -0,0 +1,41 @@
+import { bytes, io, fmt, runtime, strings }
+
+def Main(start any) (stop any) {
+	startFanOut FanOut<any>
```
emil14: first of all nodes names must be lower_snake_case, that's stated in the docs/style_guide.md (I wish it would be just docs/style.md btw)

second - Im not sure if we have this rule, but components should be rather simple, I recall adding such rule but dont see it in the docs, we should add one, this one for intance is way to big, that's probably the case for all the neva components in this and perhaps other repos

finally it seems like either the skill neva authoring was not used or it was ignored for some reason, would be nice to figure out whuy

internal/runtime/funcs/utils.go
```diff
@@ -23,6 +23,10 @@ func errFromString(s string) runtime.StructMsg {
 	})
 }
 
+func sendRuntimeError(ctx context.Context, errOut runtime.SingleOutport, err error) bool {
```
emil14: why this secion

```md
## Comments

- Add doc comments for new Go functions and types.
- For newly generated Go code blocks longer than 3 lines, add a short one-line intent comment when the purpose is not immediately obvious.
```

was ignored from skill?

internal/runtime/funcs/utils.go
```diff
@@ -23,6 +23,10 @@ func errFromString(s string) runtime.StructMsg {
...
+	return errOut.Send(ctx, errFromErr(err))
```
emil14: is `errFromErr(err)` all we abstract away with this? add rule to go authoring skill that we should only create functions/methods if we abstract away at least two things, this one is useless for instance

internal/runtime/funcs/registry.go
```diff
@@ -8,6 +8,8 @@ import (
 
 //nolint:funlen // TODO(strict-lint phase 1): temporary suppression; remove after strict cleanup.
 func NewRegistry() map[string]runtime.FuncCreator {
+	fileHandles := newFileHandleStore()
```
emil14: would be nice to see why it's created exactly here and what is it at all, looks confusing

internal/runtime/funcs/file_handles.go
```diff
@@ -0,0 +1,97 @@
...
+	"fmt"
+	"os"
+	"sync"
+
+	"github.com/nevalang/neva/internal/runtime"
+)
+
+//nolint:govet // Keeping map and counter together improves readability for this small state holder.
+type fileHandleStore struct {
+	mu     sync.Mutex
```
emil14: make sure this won't be bottleneck for the runtime

internal/runtime/funcs/file_handles.go
```diff
@@ -0,0 +1,97 @@
...
+	nextID int64
+	files  map[int64]*os.File
+	stdio  map[int64]struct{}
```
emil14: every field needs a comment

internal/runtime/funcs/file_close.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/nevalang/neva/internal/runtime"
+)
+
+type fileClose struct {
+	handles *fileHandleStore
```
emil14: wdyt about passing this to context? or at all passing from the runtime down instead of creating inside func registry?

just asking

internal/runtime/funcs/file_handles.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+const (
```
emil14: is this how Go does it? We should mostly mimic go

internal/runtime/funcs/file_handles.go
```diff
@@ -0,0 +1,97 @@
...
+
+func newFileHandleStore() *fileHandleStore {
+	files := map[int64]*os.File{
+		stdinFileHandleID:  os.Stdin,
+		stdoutFileHandleID: os.Stdout,
+		stderrFileHandleID: os.Stderr,
+	}
+
+	return &fileHandleStore{
+		nextID: 3,
```
emil14: replace with stderrFileHandleID+1 for clarity

internal/runtime/funcs/file_handles.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+const (
+	stdinFileHandleID  int64 = 0
+	stdoutFileHandleID int64 = 1
+	stderrFileHandleID int64 = 2
+)
+
+func newFileHandleStore() *fileHandleStore {
+	files := map[int64]*os.File{
```
emil14: do we collect garbage anyhow? or do we defer that for the future? will this map grow forever? is Close method cleans it?

internal/runtime/funcs/file_handles.go
```diff
@@ -0,0 +1,97 @@
...
+		stdinFileHandleID:  os.Stdin,
+		stdoutFileHandleID: os.Stdout,
+		stderrFileHandleID: os.Stderr,
+	}
+
+	return &fileHandleStore{
+		nextID: 3,
+		files:  files,
+		stdio: map[int64]struct{}{
```
emil14: so it's a set? hard to see the purpose from the glance

internal/runtime/funcs/file_handles.go
```diff
@@ -0,0 +1,97 @@
...
+			stdinFileHandleID:  {},
+			stdoutFileHandleID: {},
+			stderrFileHandleID: {},
+		},
+	}
+}
+
+func (s *fileHandleStore) Add(file *os.File) int64 {
+	s.mu.Lock()
```
emil14: so it's a global bottheneck huh?

internal/runtime/funcs/file_open.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/nevalang/neva/internal/runtime"
+)
+
+type fileOpen struct {
+	handles *fileHandleStore
```
emil14: Should this be part of the runtime API maybe? Im not sure if I like how it looks now...

internal/runtime/funcs/file_read_all.go
```diff
@@ -48,6 +48,7 @@ func (c fileReadAll) Create(rio runtime.IO, _ runtime.Msg) (func(ctx context.Con
 
 			data, err := io.ReadAll(f)
 			if err != nil {
+				_ = f.Close()
```
emil14: one more thing for go authoring skill if its still not there is to NEVER ignore errors

examples/file_handles/main.neva
emil14: Fixed across 36b26fa8 and 0bdeb5c6:

- renamed the example nodes to lower_snake_case
- split the large graph into Main/App/WriteExample/ReadExample
- added the small/focused component rule to docs/style_guide.md

On the skill point: this PR branch does not contain .codex/skills; it has .claude/rules. I followed the Neva style guide/rule file in this checkout and added the missing durable style rule to docs.

internal/runtime/funcs/utils.go
emil14: Fixed in 36b26fa8 by deleting sendRuntimeError. There is no new helper left here that needs a doc comment.

internal/runtime/funcs/utils.go
emil14: Fixed in 36b26fa8: removed this helper and inlined errOut.Send(ctx, errFromErr(err)). I also added the abstraction rule to the local Go authoring rule file present on this branch (.claude/rules/go-files.md). This branch does not have .codex/skills yet, so that source may need to be ported separately if main has moved the rules.

internal/runtime/funcs/registry.go
emil14: Fixed in 5a371b12: added a short comment explaining that the store is created once per registry so Open/Create/ReadAllFile/WriteAllFile/Close nodes share handle identity.

internal/runtime/funcs/file_handles.go
emil14: I checked the critical section: the mutex is held only for handle metadata operations (add/get/delete/stdout guard). Actual Read/Write/Close I/O happens outside the lock, so normal file I/O is not serialized through this store.

This can still become a contention point for workloads that churn handles heavily. Options if we want to harden that now: use RWMutex for Get-heavy flows, use sync.Map + atomic IDs, or move handles into a runtime-owned table. I would keep the simple mutex for this PR unless we want to add a benchmark-driven requirement.

internal/runtime/funcs/file_handles.go
emil14: Fixed in 36b26fa8: added field comments for the mutex, next handle ID, handle map, and stdio handle set.

internal/runtime/funcs/file_close.go
emil14: I think a runtime-owned handle table could be cleaner long-term than hiding this inside funcs.NewRegistry. Passing through context feels weaker to me because it makes the dependency implicit and easy to miss.

Current registry-level storage was the smallest change that preserves handle sharing across file components. If we want to change direction, I see two concrete options: add handle storage to runtime.IO/runtime state and pass it down explicitly, or make registry construction accept shared runtime resources. Which one matches the runtime architecture you want?

internal/runtime/funcs/file_handles.go
emil14: This currently mimics POSIX fd numbering for stdio (0/1/2), not Go API shape. If we want to mostly mimic Go, then exposing File as an int-like handle is probably the bigger mismatch; Go would model this closer to an opaque *os.File-like resource, with stdio as predefined variables/resources rather than magic numeric IDs.

For this PR I kept File opaque at the Neva level but int-backed in runtime. The decision to make is whether Neva File should behave more like fd handles or more like Go files.

internal/runtime/funcs/file_handles.go
emil14: Fixed in 36b26fa8: nextID now starts at stderrFileHandleID + 1.

internal/runtime/funcs/file_handles.go
emil14: Close deletes non-stdio handles from the map before closing the underlying file, so correctly closed handles do not accumulate. Stdio handles are intentionally permanent.

There is no GC/finalizer cleanup right now: if user code opens/creates and never closes, the handle stays until the registry/program ends. We can add runtime-level cleanup later, but that depends on whether handles stay registry-local or move into runtime state.

internal/runtime/funcs/file_handles.go
emil14: Fixed in 36b26fa8: renamed stdio to stdioHandleIDs and documented it as the set of stdio-backed handles that must not be closed.

internal/runtime/funcs/file_handles.go
emil14: It is global per registry, but it is not a global I/O lock: the lock only protects metadata lookup/add/delete; reads and writes happen after Get returns and outside the mutex.

If we expect many components opening/closing/looking up handles concurrently, the next improvement would be RWMutex or sync.Map + atomic IDs. I would prefer adding that only with a contention benchmark or a runtime-level handle-table decision.

internal/runtime/funcs/file_open.go
emil14: Maybe, yes. A runtime API/resource table would make the ownership more explicit than file funcs sharing private state. I did not move it in this pass because that touches the runtime boundary and constructor APIs beyond file handles.

Concrete decision needed: should file handles be a funcs-local implementation detail for now, or should runtime.IO/runtime state expose a first-class resource table that extern funcs can use?

internal/runtime/funcs/file_read_all.go
emil14: Fixed in 36b26fa8: the ReadAll error path now checks the close error and reports it together with the read error. I also added the NEVER-ignore-errors rule to the local Go authoring rule file on this branch.

internal/runtime/funcs/utils.go
emil14: we will merge actual fresh main here and you should update harness

internal/runtime/funcs/file_close.go
emil14: IDK decide for yourself, but try to make it elegant, we have enough time. Let's say context is out of the conversation. Do we already have some runtime apis that funcs use? Like something for Panic for example?

internal/runtime/funcs/file_handles.go
emil14: Remember that in Go we can pass code (functions) around, in Neva data is data (messages) and code is code (components). It means we cannot have os.File as an interface, only as a data-type, wheather that number or struct or something else. WDYT fits us best?

internal/runtime/funcs/file_handles.go
emil14: > If we expect many components opening/closing/looking up handles concurrently, the next improvement would be RWMutex or sync.Map + atomic IDs

so you sure with no benchmark specifically for that we should not do that? Isnt its like adding RWMutex will 100% help?

Also you say lock per registry, so isnt registry global? Is it like sharder or something?

internal/runtime/funcs/file_open.go
emil14: runtime.IO is too generic I guess but as `runtime` package API maybe

internal/runtime/funcs/file_read_all.go
emil14: we going to merge main so keep this up to date

.codex/skills/go-authoring/SKILL.md
```diff
@@ -20,6 +20,11 @@ Use this skill for changes touching `*.go`.
   unavoidable, keep it narrowly scoped to the exact line/rule and add a short
   justification.
 - Prefer `Makefile` targets for standard checks when applicable.
+- Do not add helper functions or methods unless they abstract at least two
+  meaningful operations or remove real complexity.
+- Never ignore errors. If an error cannot be returned or sent through the
+  component API, handle it explicitly and document or panic only when it
```
emil14: do not talk about neva here, describe simple go practices

we either handle error or we pass it

e.g. return (maybe wrap also) OR log OR panic but not both

most importantly NEVER ignore, never `_ := ...`

docs/style_guide.md
```diff
@@ -58,6 +58,8 @@ Names should inherit context from parent scope. Good naming eliminates need for
 
 ## Networks
```
emil14, line 59: I feel like this whole file could use some help with some simple and small examples since this gonna be used for AI harness

docs/style_guide.md
```diff
@@ -58,6 +58,8 @@ Names should inherit context from parent scope. Good naming eliminates need for
...
 
+- Keep components small and focused. Split large graphs into named components
+  when that makes the main dataflow easier to scan.
```
emil14: we need some numbers of connectins and/or nodes used

internal/runtime/funcs/file_handles.go
```diff
@@ -0,0 +1,96 @@
...
+import (
+	"errors"
+	"fmt"
+	"os"
+	"sync"
+
+	"github.com/nevalang/neva/internal/runtime"
+)
+
+type fileHandleStore struct {
```
emil14: type has still no comments

fields still has no comments too

emil14 (commented):
Все ок.

internal/runtime/funcs/utils.go
emil14: Merged fresh main is reflected on this branch now. `.codex/skills/go-authoring/SKILL.md` exists here, and 64d87e601b31b8ce844fc8b79e6a4ef38c56b29a keeps the harness rule up to date in the Codex skill instead of the old local rule file.

internal/runtime/funcs/file_close.go
emil14: I inspected the current runtime APIs. Funcs already depend on `runtime.IO` for ports, `runtime.TracerFromIO` for panic trace rendering, and `runtime.Terminate` for `runtime.Panic`; there is not a general runtime resource API yet.

My recommendation: do not put file handles into `context` or `runtime.IO`. If we promote this later, add an explicit runtime package resource API/resource table and pass it through registry construction. That is a cleaner direction, but it touches runtime lifecycle/constructor boundaries beyond this file API PR, so I kept this PR scoped and documented/hardened the current table in 64d87e601b31b8ce844fc8b79e6a4ef38c56b29a.

internal/runtime/funcs/file_handles.go
emil14: Given Neva data/code separation, I think the best current shape is a nominal data type in stdlib (`io.File`) backed by an opaque integer runtime handle. It keeps File as data that can move through ports, avoids pretending users can call methods/functions on it, and still prevents ordinary API surfaces from exposing raw `int` semantically.

I would not use a struct unless it carries meaningful stable user-visible fields; otherwise it invites field-level coupling to runtime internals. Long term, if the runtime gains truly opaque resource messages, `io.File` could move from int-backed to resource-token-backed without changing the user-facing component model.

internal/runtime/funcs/file_handles.go
emil14: I applied the small `RWMutex` hardening in 64d87e601b31b8ce844fc8b79e6a4ef38c56b29a: `Get` now uses `RLock`, while `Add`/`Close` still take the write lock.

It is not a 100% win for every workload: it helps concurrent read/lookups of already-open handles, but open/close churn still serializes and RWMutex can add a little overhead under write-heavy contention. It is small enough to take now.

On registry scope: `NewRegistry()` creates one handle store and closes it over the file funcs in that returned registry map. It is not a package-level global and not sharded; it is per registry instance, shared by all file funcs created from that registry.

internal/runtime/funcs/file_open.go
emil14: Agreed that `runtime.IO` is too generic for this. If we move this out of funcs-local state, I would make it a separate runtime package API, something like an explicit resource/file-handle table owned by runtime setup and injected into funcs/registry construction.

I did not fold that into this PR because it is a broader runtime lifecycle API change, not just file handles. For this PR I kept the store private to the registry and documented/hardened it in 64d87e601b31b8ce844fc8b79e6a4ef38c56b29a.

internal/runtime/funcs/file_read_all.go
emil14: Fresh main is merged into this branch now and the rule now lives in `.codex/skills/go-authoring/SKILL.md`. 64d87e601b31b8ce844fc8b79e6a4ef38c56b29a keeps the never-ignore-errors guidance in that Codex skill.

.codex/skills/go-authoring/SKILL.md
emil14: Fixed in 64d87e601b31b8ce844fc8b79e6a4ef38c56b29a. The rule now describes plain Go error handling: never ignore errors or assign them to `_`; either handle locally or pass to the caller; choose one reporting path such as return/wrap, log, or panic.

docs/style_guide.md
emil14: Added a small focused network example in 64d87e601b31b8ce844fc8b79e6a4ef38c56b29a so this section is more useful as AI harness guidance without turning the whole style guide into a larger docs pass.

docs/style_guide.md
emil14: Added concrete heuristic numbers in 64d87e601b31b8ce844fc8b79e6a4ef38c56b29a: consider splitting above 7 nodes or 12 connections unless there is a clear reason to stay flat; 3-5 nodes and up to 8 connections is the preferred readable range.

internal/runtime/funcs/file_handles.go
emil14: Fixed in 64d87e601b31b8ce844fc8b79e6a4ef38c56b29a. `fileHandleStore` now has a type comment and each field has an explanatory comment.

internal/runtime/funcs/file_handles.go
emil14: Updated in 64d87e601b31b8ce844fc8b79e6a4ef38c56b29a: the store now uses `RWMutex`, so concurrent `Get` calls no longer block each other. The store still only protects handle metadata; actual file I/O remains outside the lock.

internal/runtime/funcs/file_handles.go
emil14: I mean its a single instance shared by the whole Neva program isn't it?

internal/runtime/funcs/file_handles.go
emil14: So is it how it works now or we should change something?

internal/runtime/funcs/file_close.go
emil14: It's completely ok to touch `runtime` package, we are not limited by `runtime/funcs` scope. We must do what best for the language. I think it's ok to add type with some methods that is performant (i.e. is not program-wise bottleneck) that can be called by the runtime funcs, maybe it can implement sharding under the hood (IDK if thats how we solve bottleneck if we have it, we need to brainstorm this)

emil14: Yes, I think touching `runtime` is the right direction if we want to improve the architecture rather than just patch the current PR shape.

My recommendation is still to keep this PR as the minimal file-handle feature and do the runtime refactor as a follow-up: introduce an explicit runtime-owned resource table API, let it own file-handle lifecycle, and inject it into registry construction. That gives us one place to hide implementation details like sharding, cleanup, lock strategy, and future non-file resources.

I would avoid putting methods on the user-visible `io.File` message itself. The better split is: `io.File` stays a data token in the language, while runtime owns the operational table behind that token.

internal/runtime/funcs/file_handles.go
emil14: For this PR, I would keep it as it works now.

The user-facing shape should stay `io.File` as a nominal data type, with the runtime backing remaining opaque. I would not change it to a struct just to mirror implementation details. The change I would make later is underneath that surface: move the backing store from funcs-local state into an explicit runtime resource API, while keeping the same `io.File` component model in Neva.

internal/runtime/funcs/file_handles.go
emil14: In the current execution model, yes: effectively one store per program invocation.

More precisely, the generated entrypoints call `funcs.NewRegistry()` once per `runtime.Run(...)` or `runtime.Call(...)`, and that one registry closes over one file-handle store shared by all file funcs in that invocation. So it is not a package-global singleton across the whole process, but for one Neva program run it is a single shared instance unless the host creates multiple separate runtime calls/registries.

internal/runtime/funcs/file_open.go
emil14: Stop wrestling with me and do what I say. We will change runtime package in this PR.

internal/runtime/funcs/file_close.go
emil14: IDK what you talking about "I would avoid putting methods on the user-visible io.File message itself."

There are no messages on io.File in Neva there are no methods in general. I was talking about adding code to `runtime` package. And we WILL do it

internal/runtime/funcs/file_handles.go
emil14: I dont care if it's not literally pkg-level global variable, I care that its a progral-wise bottleneck which you trying not to admit. We MUST fix before merging

internal/runtime/funcs/file_close.go
emil14: Implemented in 997e8109ee10e6c3ee9e2695bacce3304bd1198f.

The file-handle table is now `runtime.FileHandles` in the `runtime` package, not a private `runtime/funcs` structure. The funcs layer receives `*runtime.FileHandles` from registry construction and only adapts Neva messages to runtime handle IDs. The table also owns stdio IDs and lifecycle methods (`Add`, `Get`, `Close`).

For the bottleneck concern, `runtime.FileHandles` now allocates IDs atomically and shards the handle map across 32 shard locks. `Get` uses per-shard `RLock`; `Add`/`Close` only lock the target shard. File I/O still happens outside the table locks.

internal/runtime/funcs/file_handles.go
emil14: Fixed in 997e8109ee10e6c3ee9e2695bacce3304bd1198f.

You were right about the program-wise shared instance. The new implementation removes the single table lock from that shared instance: `runtime.FileHandles` uses atomic ID allocation plus 32 shards, so independent handles do not contend on one program-wide mutex for lookup/add/close metadata. This still preserves one logical handle table per registry/program invocation, but avoids one global metadata lock for the whole program.

internal/runtime/funcs/file_open.go
emil14: Implemented in 997e8109ee10e6c3ee9e2695bacce3304bd1198f.

The API is now in `runtime`: `runtime.NewFileHandles()`, `runtime.FileHandles`, and stdio handle constants. `funcs.NewRegistry()` creates that runtime table and injects it into the file funcs it registers, which fits the current architecture where callers pass a registry into `runtime.Run`/`runtime.Call`.

So this PR now changes the runtime package directly while keeping registry construction as the place where the per-program table is wired into extern funcs.

docs/style_guide.md
```diff
@@ -58,9 +58,32 @@ Names should inherit context from parent scope. Good naming eliminates need for
...
 - Use `?` to propagate errors unless custom error handling is needed.
-- Prefer chaining connections inline when possible (e.g. `c -> switch:case[0] -> fmt.Println`) to keep the dataflow compact and easier to scan.
+- Prefer chaining connections inline when possible
+  (e.g. `c -> switch:case[0] -> fmt.Println`) to keep the dataflow compact and
+  easier to scan.
+
+Small component example (assuming `bytes` and `io` are imported):
+
+```neva
+def SaveGreeting(start any, text string) (stop any, err error) {
```
emil14: 1. this is invalid example because :start port referenced more than once, for fan-out -> [] syntax must be used
2. should be sig, text instead of start, text and res, err instead of stop, err (port naming conventions should be part of this doc)
3. this doc overall is rather short, your text is way too descriptive

docs/style_guide.md
```diff
@@ -58,9 +58,32 @@ Names should inherit context from parent scope. Good naming eliminates need for
...
+	bytes_from_string bytes.FromString
+	write io.WriteAll
+	---
+	:text -> bytes_from_string
+	bytes_from_string -> write:data
+	:start -> 'greeting.txt' -> write:filename
```
emil14: it's much better to just ref the only connection intself no need for the whole component

docs/style_guide.md
```diff
@@ -58,9 +58,32 @@ Names should inherit context from parent scope. Good naming eliminates need for
 
 ## Networks
 
-- Omit port names when possible. It enables renaming of ports without updating the networks.
+- Keep components small and focused. Split large graphs into named components
+  when that makes the main dataflow easier to scan. As a rule of thumb, a
+  component with more than 7 nodes or 12 connections should either have a clear
```
emil14: 7 nodes and 12 connections is close to unreadable

5 nodes and 10 connections is a hard limit, 3 nodes with 5 connections is the perfect size

of course there can be exceptions like in any rule

internal/runtime/funcs/file_close.go
emil14: Adjusted in 222ad4b9d33d3c82020f2381c97840c86e41b092.

I removed the over-engineered sharding/resource-table shape. What remains is the intended simpler runtime package API: `runtime.FileHandles` is a plain thread-safe struct with a map, stdio handle set, counter, `RWMutex`, and `Add/Get/Close` methods. `runtime/funcs` no longer owns that structure; it only receives `*runtime.FileHandles` from registry construction and uses the runtime methods.

internal/runtime/funcs/file_handles.go
emil14: Adjusted in 222ad4b9d33d3c82020f2381c97840c86e41b092.

I removed sharding here. The implementation is now the straightforward thread-safe runtime struct: a map plus `RWMutex`. That keeps the design small and explicit while still moving ownership into `runtime`, which was the main architectural problem.

internal/runtime/funcs/file_open.go
emil14: Implemented and then simplified in 222ad4b9d33d3c82020f2381c97840c86e41b092.

The runtime package now owns the public-ish internal API (`runtime.FileHandles`, `runtime.NewFileHandles`, stdio handle constants). I removed the extra sharding/resource-manager complexity; this is now just the runtime-level type with methods that file funcs call.

internal/runtime/file_handles.go
```diff
@@ -0,0 +1,115 @@
...
+	handles.addStdio(StdinFileHandleID, os.Stdin)
+	handles.addStdio(StdoutFileHandleID, os.Stdout)
+	handles.addStdio(StderrFileHandleID, os.Stderr)
+
+	return handles
+}
+
+// Add stores file and returns a new opaque runtime handle ID.
+func (handles *FileHandles) Add(file *os.File) int64 {
+	handleID := handles.nextID.Add(1) - 1
```
emil14: so we every time choose the next shard? the counter is increased on every call to Add()?

internal/runtime/file_handles.go
```diff
@@ -0,0 +1,115 @@
...
+	defer shard.mu.Unlock()
+
+	shard.files[handleID] = file
+
+	return handleID
+}
+
+// Get returns the file registered for handleID.
+func (handles *FileHandles) Get(handleID int64) (*os.File, error) {
+	shard := handles.shard(handleID)
```
emil14: handleID is what components (runtime funcs) pass around as file descriptor number?

internal/runtime/file_handles.go
```diff
@@ -0,0 +1,115 @@
...
+		return nil, fmt.Errorf("file handle %d not found", handleID)
+	}
+
+	return file, nil
+}
+
+// Close removes and closes a dynamic file handle.
+func (handles *FileHandles) Close(handleID int64) error {
+	if isStdioFileHandleID(handleID) {
+		return fmt.Errorf("cannot close stdio file handle %d", handleID)
```
emil14, line 107: shouldn't we maybe panic from here? does it make sense to return to component?

Im not sure just asking

internal/runtime/file_handles.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	shard := handles.shard(handleID)
+
+	shard.mu.Lock()
+	file, found := shard.files[handleID]
+	if found {
+		delete(shard.files, handleID)
```
emil14: I guess we can call Close() on the same handleID many times, so it's idempotent? would be nice to documentthat

internal/runtime/funcs/file_create.go
```diff
@@ -0,0 +1,66 @@
...
+		return errOut.Send(ctx, errFromErr(err))
+	}
+
+	handleID := c.handles.Add(file)
+	if resOut.Send(ctx, runtime.NewIntMsg(handleID)) {
+		return true
+	}
+
+	if err := c.handles.Close(handleID); err != nil {
+		panic(err)
```
emil14: I personally find it weird that we panic from component while processing the message...

maybe we should do that from runtime OR return err from this function? Not sure...

Also most importantly I do not understand, we create file, we open it and we close it right away?

internal/runtime/funcs/file_handles.go
```diff
@@ -0,0 +1,16 @@
+package funcs
+
+import (
+	"errors"
+
+	"github.com/nevalang/neva/internal/runtime"
+)
+
+func fileHandleID(msg runtime.Msg) (int64, error) {
```
emil14: should this be in utils.go?

internal/runtime/funcs/file_open.go
```diff
@@ -0,0 +1,65 @@
...
+			}
+
+			if !c.handleFileMessage(ctx, nameMsg, resOut, errOut) {
+				return
+			}
+		}
+	}, nil
+}
+
+func (c fileOpen) handleFileMessage(
```
emil14, line 45: this seems to be a boilerplate no? maybe move to utils?

internal/runtime/funcs/file_read_all_handle.go
```diff
@@ -0,0 +1,80 @@
...
+
+	data, err := io.ReadAll(file)
+	if err != nil {
+		if !handleOut.Send(ctx, runtime.NewIntMsg(handleID)) {
+			return false
+		}
+		return errOut.Send(ctx, errFromErr(err))
+	}
+
+	if !resOut.Send(ctx, runtime.NewBytesMsg(data)) {
```
emil14, line 75: so read all should not close?

internal/runtime/funcs/file_stdio.go
```diff
@@ -0,0 +1,86 @@
...
+			if _, ok := sigIn.Receive(ctx); !ok {
+				return
+			}
+
+			if !resOut.Send(ctx, runtime.NewIntMsg(runtime.StderrFileHandleID)) {
+				return
+			}
+		}
+	}, nil
+}
```
emil14, line 86: absolutely every runtime function MUST have BOTH unit tests AND benchmarks

internal/runtime/funcs/file_write_all_handle.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	if _, err := file.Write(dataMsg.Bytes()); err != nil {
+		if !resOut.Send(ctx, runtime.NewIntMsg(handleID)) {
+			return false
+		}
+		return errOut.Send(ctx, errFromErr(err))
+	}
+
+	return resOut.Send(ctx, runtime.NewIntMsg(handleID))
```
emil14, line 73: so also no close for this?

docs/style_guide.md
emil14: Fixed in 9ce00f7d60bc96875b5fd144958454604ba4847a. I removed the invalid full component example, replaced it with a single connection example, shortened the wording, and added the standard flow-name rule for sig/res/err.

docs/style_guide.md
emil14: Fixed in 9ce00f7d60bc96875b5fd144958454604ba4847a. The example is now just one connection, not a whole component.

docs/style_guide.md
emil14: Fixed in 9ce00f7d60bc96875b5fd144958454604ba4847a. The rule now says to aim for about 3 nodes and 5 connections, and to split at 5 nodes or 10 connections unless the flat graph has a clear reason to stay together.

internal/runtime/file_handles.go
emil14: This is stale after the simplification. Sharding was removed in 222ad4b9; Add now takes the FileHandles mutex, uses nextID as the new opaque handle ID, increments nextID once, and stores the file in the map.

internal/runtime/file_handles.go
emil14: Yes, runtime funcs pass this handleID around as the Neva-visible file token. It is deliberately not documented as an OS file descriptor: stdio mirrors 0/1/2, but dynamic handles are opaque runtime IDs owned by runtime.FileHandles.

internal/runtime/file_handles.go
emil14: I kept this as an error, not a panic. Unknown/already-closed/stdio-close handles are runtime operation errors that the file_close component can send on err; panic should stay for internal impossible states. That also keeps the behavior consistent with other file funcs: invalid user/runtime data flows through err.

internal/runtime/file_handles.go
emil14: Documented in 9ce00f7d60bc96875b5fd144958454604ba4847a, but with the opposite contract: Close is not idempotent. The first Close removes and closes the dynamic handle; a second Close returns an unknown/already-closed handle error so user code can detect double-close bugs.

internal/runtime/funcs/file_create.go
emil14: Fixed in 9ce00f7d60bc96875b5fd144958454604ba4847a. Open/Create no longer panic in the component cleanup path. They keep the file open on the successful path and pass the handle downstream. The only close right after open/create is the failure-to-deliver cleanup path, so we do not leak a newly opened file if resOut.Send returns false.

internal/runtime/funcs/file_handles.go
emil14: I kept this in file_handles.go rather than utils.go because it is file-API-specific, not a generic funcs utility. The new shared open/create helper also lives there, so the file-specific adapters are grouped in one place.

internal/runtime/funcs/file_open.go
emil14: Fixed in 9ce00f7d60bc96875b5fd144958454604ba4847a. The open/create boilerplate is now sendFileHandle in file_handles.go. I used file_handles.go instead of utils.go because the helper is specifically about runtime file-handle ownership and cleanup.

internal/runtime/funcs/file_read_all_handle.go
emil14: Correct: handle-based ReadAllFile should not close. It reads from the current file handle and returns the same handle so the caller can keep using it or close it explicitly. 9ce00f7d60bc96875b5fd144958454604ba4847a adds a unit test for the success path that asserts both data and handle are returned.

internal/runtime/funcs/file_stdio.go
emil14: Fixed in 9ce00f7d60bc96875b5fd144958454604ba4847a. I added unit coverage for file_open, file_create, file_close, file_read_all, file_write_all, file_stdin, file_stdout, and file_stderr, plus BenchmarkFile* benchmarks for the same runtime function set. Verified with go test ./internal/runtime ./internal/runtime/funcs -count=1, benchmark smoke, and scoped golangci-lint.

internal/runtime/funcs/file_write_all_handle.go
emil14: Correct: handle-based WriteAllFile also should not close. It writes to the current file handle and returns the same handle, leaving lifecycle ownership with the caller. 9ce00f7d60bc96875b5fd144958454604ba4847a adds a unit test that verifies the handle remains valid after write.

docs/style_guide.md
```diff
@@ -58,9 +58,23 @@ Names should inherit context from parent scope. Good naming eliminates need for
...
+- Prefer chaining connections inline when possible
+  (e.g. `c -> switch:case[0] -> fmt.Println`) to keep the dataflow compact and
+  easier to scan.
+- Prefer standard flow names: `sig` for trigger inputs, `res` for success
+  outputs, and `err` for errors.
+
+Example:
+
+```neva
+read:res -> strings.FromBytes -> :res
```
emil14: So, there is a problem here, at least in this file, and maybe somewhere else, I'm not sure. With the way you refer to the node inside the connection, so it's not, and it can't be a strings.fromBytes, because that is a reference to component, not to the node, right? This is the problem that we got all the time. Even though we have a harness to check, like, nivacode, I don't know what to do about that, maybe you have some idea, but right now in here, we have to fix that, right? So here on line 76 and above on a line 68, the node name should be, like, lowercase, from bytes with lowerSnakeCase, and on the 68 lines, gotta be just a print line, lowercase

internal/runtime/funcs/file_bench_test.go
```diff
@@ -0,0 +1,205 @@
+package funcs
```
emil14: so the name of the benchmark have to be the same as name of the file with the actual code of runtime function and I'm not sure if that's the case here so I want you to check that

internal/runtime/file_handles.go
```diff
@@ -0,0 +1,95 @@
...
+	StdoutFileHandleID int64 = 1
+	StderrFileHandleID int64 = 2
+)
+
+// FileHandles owns runtime file resources exposed to Neva as opaque integer handles.
+type FileHandles struct {
+	// files maps Neva-visible handle IDs to live process files.
+	files map[int64]*os.File
+	// stdioHandleIDs marks process stdio handles that must not be closed by Neva.
+	stdioHandleIDs map[int64]struct{}
```
emil14, line 35: I want you to elaborate on that like what's the case with it why there are some handles that Niva must not close and who closes them I feel like we could use some more explanation on what's going on here

internal/runtime/file_handles.go
```diff
@@ -0,0 +1,95 @@
...
+
+// Stdio file handle IDs mirror the conventional process file descriptors.
+const (
+	StdinFileHandleID  int64 = 0
+	StdoutFileHandleID int64 = 1
+	StderrFileHandleID int64 = 2
+)
+
+// FileHandles owns runtime file resources exposed to Neva as opaque integer handles.
+type FileHandles struct {
```
emil14, line 29: So, correct me if I'm wrong but from the comments in this political quest like from the discussions I feel like we did have some kind of sharding to avoid having a bottleneck and now like you said that we're not sharding anymore and how that exactly is working at the moment like are we having just a single instance of file handles so every like file operation is a mutex log that blocks basically the whole program and I'm not saying that like in theory we could have like multiple file handle instance I'm talking about how that works in practice how that actually works at the moment you know because if that's a bottleneck that's like that's a problem and maybe we could leave with that but then we have to create like an issue for that but most of all first of all we I must have a clear understanding of what's actually going on at the moment

internal/runtime/file_handles.go
```diff
@@ -0,0 +1,95 @@
...
+	// files maps Neva-visible handle IDs to live process files.
+	files map[int64]*os.File
+	// stdioHandleIDs marks process stdio handles that must not be closed by Neva.
+	stdioHandleIDs map[int64]struct{}
+	// nextID is the next dynamic handle ID returned by Add.
+	nextID int64
```
emil14, line 39: and again we could use more explanation here so like are we talking about some kind of round robin here when I look at this it's like I have no idea why would we want to have like a dynamic ID and some kind of like a cursor or whatever

internal/runtime/file_handles.go
```diff
@@ -0,0 +1,95 @@
...
+	// mu protects files and nextID.
+	mu sync.RWMutex
```
emil14, line 42: same we could use help from couple more lines of comments like why is it on RV mutics what does it mean and so on and so forth

internal/runtime/file_handles.go
```diff
@@ -0,0 +1,95 @@
...
+			StdoutFileHandleID: {},
+			StderrFileHandleID: {},
+		},
+	}
+}
+
+// Add stores file and returns a new opaque runtime handle ID.
+func (handles *FileHandles) Add(file *os.File) int64 {
+	handles.mu.Lock()
+	defer handles.mu.Unlock()
```
emil14, line 70: so the MU here is like an RWMutex which means like read write if I'm not mistaken so here we call it like lock and unlock which means we block both like reading and writing operations at the same time and I'm I think we could use help of something like little bit more comments on why is it like this because I guess we have to block reading operations when we write but not writing operations when we read it. It doesn't make sense at the moment like I mean in the Gath we're only blocking the reading operations I mean is it really correct could you please explain why is it

internal/runtime/file_handles.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	file, found := handles.files[handleID]
+	if found {
+		delete(handles.files, handleID)
+	}
+	handles.mu.Unlock()
+
+	if !found {
+		return fmt.Errorf("file handle %d not found", handleID)
```
emil14, line 117: so if it's not found we just do nothing which oh no actually I see I see we are my question is how is that working in go and like shouldn't that be like it in potent operation like wouldn't it be better to allow user to close multiple times without theory of anything wrong you know what's the position in good programming languages on that I would say or even better like the programming community in general and in the you know programming languages and compiler is design and implementation like creators what people think about that generally shouldn't it be an impotent and how that shouldn't it be an impotent in general and what state what's goes take on that specifically and is go is a good example for this

internal/runtime/funcs/file_handles.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"context"
+	"errors"
+	"os"
+
+	"github.com/nevalang/neva/internal/runtime"
+)
+
+func fileHandleID(msg runtime.Msg) (int64, error) {
```
emil14: every fucking new generated function must have a like comment like .com and explaining the heck is doing and why so here for instance I assume we're like not expecting this function to be called with something that is not an integer what this function is actually doing is just like casting it to I think it's a little bit misleading like it just tries to get the integer from it I don't really know maybe it's not a big deal but I don't know man maybe we should move that to your teals and assign like a little bit more a generic name honestly in my opinion like all this kind of code should anyway go to your teals that could be like a big file but we could we could have a sections or maybe it could be like your teals underscore file underscore handles I don't know the problem with the actual naming is that it feels like this is a runtime function which is not actually

internal/runtime/funcs/file_handles.go
```diff
@@ -0,0 +1,37 @@
...
+	idMsg, isIntMsg := msg.(runtime.IntMsg)
+	if !isIntMsg {
+		return 0, errors.New("file handle must be int")
+	}
+
+	return idMsg.Int(), nil
+}
+
+// sendFileHandle stores file, sends its handle, and closes it if delivery fails.
+func sendFileHandle(
```
emil14: from the description of this function it feels a little bit complicated and we could use a bit more comments on why does it exists what for who's using it and so on and so forth because for me it's a little bit not obvious I assume that's some kind of pattern that like arises arises multiple times and run a functions but still I don't want to be guessing you know

internal/runtime/funcs/file_create.go
```diff
@@ -0,0 +1,58 @@
...
+	resOut runtime.SingleOutport,
+	errOut runtime.SingleOutport,
+) bool {
+	// #nosec G304 -- filename is user-controlled by design.
+	file, err := os.OpenFile(nameMsg.Str(), os.O_CREATE|os.O_RDWR|os.O_TRUNC, 0o644)
+	if err != nil {
+		return errOut.Send(ctx, errFromErr(err))
+	}
+
+	return sendFileHandle(ctx, c.handles, file, resOut, errOut)
```
emil14: yeah like send file handle is a little bit more complicated than it feels like based on just its name now I kind of understand like what it does it takes a file which is a go abstraction if I'm not mistaken an interface actually and does something with it like stores the actual file in memory and takes the like ID cursor of that file and sends it further to the result output and sends the ear sense the ear if it can do that to the ear out port for some reason also why like is it like send file handle if I'm not mistaken trying to also click close the file if something goes wrong I'm not fucking remembering all all the things but we could use like more comments in more descriptive names for sure

docs/style_guide.md
emil14: Fixed in 66d50434. The style guide examples now use node instance names rather than component references: `println` in the inline chain example and `fromBytes` in the standalone connection example. I agree this is exactly the kind of snippet mistake the Neva-code harness should catch more systematically; for this PR I fixed the concrete docs issue.

internal/runtime/funcs/file_bench_test.go
emil14: Fixed in 66d50434. I split the generic `file_bench_test.go` into benchmark files that match the runtime function files: `file_open_bench_test.go`, `file_create_bench_test.go`, `file_close_bench_test.go`, `file_read_all_handle_bench_test.go`, `file_write_all_handle_bench_test.go`, and `file_stdio_bench_test.go`. Shared benchmark-only helpers live in `file_bench_helpers_test.go`.

internal/runtime/file_handles.go
emil14: Fixed in 66d50434. I expanded the comments around `stdioHandleIDs`: stdio handles are owned by the host process, so Neva can use them but must not close them. Their lifetime is managed outside the Neva file-handle table, unlike dynamic handles returned by file_open/file_create.

internal/runtime/file_handles.go
emil14: Fixed in 66d50434 with a longer `FileHandles` type comment. Current behavior is one shared `FileHandles` table per stdlib registry, but the mutex protects only the map/ID bookkeeping. `Get` takes `RLock` for a short lookup and releases it before actual `ReadAll`/`Write` I/O, so normal I/O through already-open handles is not serialized by this table. `Add`/`Close` still serialize because they mutate the map and `nextID`. I kept sharding out for now because the current bottleneck is only open/close bookkeeping, not whole-program file I/O; if open/close churn becomes measured pain, sharding can be added inside `runtime.FileHandles` without changing the public runtime API.

internal/runtime/file_handles.go
emil14: Fixed in 66d50434. I added a field comment explaining that `nextID` is a monotonic cursor for dynamic handles created by Open/Create, not round-robin. IDs are not reused so stale handle use remains visible as an unknown-handle error after Close removes the old entry.

internal/runtime/file_handles.go
emil14: Fixed in 66d50434. The `mu` field comment now says exactly what it protects and why it is an `RWMutex`: `Get` uses `RLock` for short concurrent map lookups, while `Add` and `Close` use `Lock` because they mutate the table.

internal/runtime/file_handles.go
emil14: Fixed in 66d50434. I added comments explaining the lock split: reads use `RLock` because multiple lookups can safely run together; writes use `Lock` because changing a Go map or advancing `nextID` must exclude both readers and other writers. Importantly, `Get` releases the lock before returning the `*os.File`, so the later file I/O does not hold the table lock.

internal/runtime/file_handles.go
emil14: Fixed in 66d50434 by documenting the intended non-idempotent contract on `Close`. A second close returns an unknown-handle error rather than silently succeeding. That makes double-close/stale-handle bugs visible to Neva code through the `err` outport, and it follows Go’s general posture that using a closed resource is an error rather than a harmless no-op.

internal/runtime/funcs/file_handles.go
emil14: Fixed in 66d50434. I renamed `fileHandleID` to `parseFileHandleID` and added a doc comment explaining that it extracts the opaque runtime file-handle ID from a Neva `int` message and centralizes the type check/error text for all handle-based file externs. I kept it in `file_handles.go` because it is file-handle-specific helper code, not a runtime function implementation.

internal/runtime/funcs/file_handles.go
emil14: Fixed in 66d50434. I renamed the helper to `storeAndSendFileHandle` and expanded its doc comment. The comment now explains why it exists: file_open and file_create both transfer an opened `*os.File` into `runtime.FileHandles`, expose only the opaque ID to Neva, and must close/remove the file immediately if the result handle cannot be delivered.

internal/runtime/funcs/file_create.go
emil14: Fixed in 66d50434. The call now uses the more explicit `storeAndSendFileHandle` name, and the helper comment spells out the full behavior: register the opened file in the runtime table, send the opaque handle ID on `res`, and close/remove the handle if sending fails because no downstream Neva code would be able to close it.

internal/runtime/funcs/file_handles.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	return idMsg.Int(), nil
+}
+
+// storeAndSendFileHandle registers an opened file and sends its handle to Neva.
+//
+// file_open and file_create both transfer the opened *os.File into
+// runtime.FileHandles, then expose only the opaque handle ID downstream. If the
+// result cannot be delivered, no Neva node can close that new handle, so this
```
emil14: > If the result cannot be delivered

This line

```neva
if resOut.Send(ctx, runtime.NewIntMsg(handleID)) {
```

means that program must terminate!

Correct me if I'm wrong but methods like `Send` or `Receive` return `bool` as an indicator if `closed==true`. 

Or is it vice versa?

PR #1041, opened 2026-02-24
Implement std/os package for env/process/fs primitives

### LINKED ISSUE #586: Implement `os` package for stdlib
The documentation page does not expose usage stats, but in everyday Go code the most commonly used `os` functions cluster around files, directories, env vars, and processes.

## Very common file helpers

- **Open / Create / OpenFile / NewFile** – opening existing files, creating new ones, or opening with specific flags and permissions.
- ReadFile / WriteFile – convenience helpers that read or write an entire file in one call.
- Stat / Lstat / FileInfo methods (from `File.Stat`, etc.) – getting file metadata like size, perms, and modification time.
- Remove / RemoveAll / Rename / Truncate – deleting files, recursively removing trees, renaming, and truncating.

## Directories and temporary paths

- Mkdir / MkdirAll – creating single directories or full directory trees.
- MkdirTemp / CreateTemp / TempDir – managing temporary directories or files for scratch work.
- ReadDir / File.ReadDir – listing directory entries in a portable way.

## Environment variables and config

- Getenv / LookupEnv – reading environment variables, with `LookupEnv` used when you need to distinguish “not set” from “empty string”.
- Setenv / Unsetenv / Clearenv / Environ – setting, removing, clearing, or enumerating environment variables.
- ExpandEnv / Expand – interpolating `$VAR` or `${VAR}` in config strings using the current environment or a custom mapping.

## Process, paths, and runtime info

- Getwd – getting the current working directory.
- Chdir – changing the current working directory in CLI tools or tests.
- Args / Exit – accessing command‑line arguments and exiting with a status code.
- Getpid / Getppid / Hostname / Executable – basic process and host information.

emil14:
## Summary
- expand `std/os` from `Args`/`Environ` to a practical typed surface for environment, process, filesystem, and temp-path operations
- add runtime function implementations for new `#extern` bindings and register them in `internal/runtime/funcs/registry.go`
- fix `args` runtime function list construction bug (pre-sized slice + append)
- add targeted unit tests for new runtime helpers and a focused e2e package for `std/os`

## Design notes
- kept this iteration value-oriented and dataflow-friendly (path/key/value/perm inputs + typed outputs)
- intentionally avoided introducing stateful file-handle API in this PR to keep semantics explicit and composable in static dataflow
- query-style components without natural data input are signal-triggered (`sig`) to avoid autonomous senders

## Validation
- `golangci-lint run ./internal/runtime/funcs/... ./e2e/os_package/...`
- `go test ./internal/runtime/funcs/... ./e2e/os_package`
- `golangci-lint run ./...`
- `go test ./...` (broad run started and progressed through many e2e packages; targeted changed-package suites fully passed)

Part of #1033
Closes #586


e2e/os_package/main/main.neva
```diff
@@ -0,0 +1,26 @@
+import {
```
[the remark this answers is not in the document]
emil14: Done: split the os package coverage into separate e2e modules for `os_args`, `os_process`, and `os_fs`, and kept `os_package` focused on env behavior.

emil14 (commented):
# Review Summary
## Critical Issues (Blocking)
- `internal/runtime/funcs/os_helpers.go:191`: `runBinaryLoop` receives from `firstIn` before it starts waiting on `secondIn`. That changes the dataflow/backpressure semantics for every new two-input `std/os` component using this helper (`Setenv`, `Mkdir`, `MkdirAll`, `Rename`, `Truncate`, `MkdirTemp`, `CreateTemp`): if the second input is produced first, or if producing the first input depends on the second branch being drained, the component can deadlock even though both inputs would eventually be available. Existing binary runtime funcs use `receive2(ctx, firstIn, secondIn)` to arm both receives concurrently. Please switch `runBinaryLoop` to `receive2` and add a regression test that proves the second input can arrive before the first without blocking.

e2e/os_args/main/main.neva
```diff
@@ -0,0 +1,16 @@
...
+
+def Main(start any) (stop any) {
+	args os.Args
+	print fmt.Println<list<string>>
+	panic runtime.Panic
+	---
+	:start -> args:sig
+	args:res -> print
+	print:res -> :stop
+	[print:err] -> panic
```
emil14: Using a single sender in multiple sender's expression, you know, like fan-in connection must be illegal because it makes no sense and only makes code a little bit more noisy. So I think that is a part, like, if that's something that is valid as of today, we should make change to compiler, to analyzer specifically, and of course add a unit test for that, but it's not. So as soon as that will happen, maybe something else will stop compiling, which is fine, which will only mean that we need to fix that. All right, but most importantly, we need to fix this, all the other occurrences, except if those are examples of non-valid code, and make change to compiler, we'll be adding unit tests. All right, I mean, it should not be legal, right? Also, don't forget, like, if a node only have one input, you should omit the name of the port, right? The same thing, by the way, is with the outports, so you could connect, like, start with arcs and print actually inside a single chain connection, which you should do. You should read the fucking style guide, and you should fucking apply that to the whole pull request. And to the other pull requests too, by the way. Thank you.

std/os/os.neva
```diff
@@ -1,13 +1,129 @@
+pub type LookupEnvResult struct {
```
emil14, line 2: my only question here basically is

- how this correlates with Go stdlib?
- is this idiomatic Neva following style guide?

e2e/os_fs/main/main.neva
```diff
@@ -0,0 +1,78 @@
+import {
+	builtin
```
emil14: please follow style guide and

- do not import builtin explicitly, maybe we should actually forbid that on compiler level, let's create issue about that
- follow style guide and split (decompose) big components (more than 3 nodes or 5 connections) into peices

not just here but this PR and other opened PRs too

e2e/os_process/main/main.neva
```diff
@@ -0,0 +1,34 @@
+import {
+	fmt
+	os
+	runtime
+}
+
+def Main(start any) (stop any) {
+	getwd os.Getwd
+	getpid os.Getpid
+	getppid os.Getppid
```
emil14: what is getppid?

internal/runtime/funcs/args_test.go
```diff
@@ -0,0 +1,18 @@
+package funcs
+
+import "testing"
+
+func TestArgsListMsg(t *testing.T) {
```
emil14, line 5: we should unit test WAY more than this, runtime functions are CRITICAL functionality, apply good table driven testing with QA best practices to cover most cases from basic to corner ones, this and other tests in this and other opened PRs, also add this to *.go rule/skill (authiring)

internal/runtime/funcs/utils_os.go
```diff
@@ -0,0 +1,214 @@
+package funcs
```
emil14, line 1: lets call it utils_os.go

internal/runtime/funcs/utils_os.go
```diff
@@ -0,0 +1,214 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/nevalang/neva/internal/runtime"
+)
+
+func singleInport(io runtime.IO, name string) (runtime.SingleInport, error) {
```
emil14, line 11: every function must have do comments

internal/runtime/funcs/utils_os.go
```diff
@@ -0,0 +1,214 @@
...
+	inport, err := io.In.Single(name)
+	if err != nil {
+		return runtime.SingleInport{}, fmt.Errorf("get inport %q: %w", name, err)
+	}
+
+	return inport, nil
+}
+
+func singleOutport(io runtime.IO, name string) (runtime.SingleOutport, error) {
+	outport, err := io.Out.Single(name)
```
emil14, line 22: why would we wanna have functions that do one thing? theres zero saving

internal/runtime/funcs/utils_os.go
```diff
@@ -0,0 +1,214 @@
...
+	step func() (runtime.Msg, error),
+) {
+	for {
+		if _, received := sigIn.Receive(ctx); !received {
+			return
+		}
+
+		result, err := step()
+		if err != nil {
+			if errOut == nil {
```
emil14, line 45: cant we just assert this once outside of the loop? also should we care at all?

e2e/os_args/main/main.neva
emil14: Fixed in b1111d0b.

- `e2e/os_args` now follows the style guide: omitted ports where inference is unambiguous and replaced `[print:err]` with the single sender form `print:err`.
- The compiler now rejects a bracketed multiple-sender expression with only one sender, e.g. `[foo] -> bar`.
- I added a parser regression test for this case.

I put the validation in the parser rather than analyzer because the parser still knows that the user wrote the bracketed multiple-sender syntax; after parsing, this would otherwise collapse to a normal one-sender connection shape.

std/os/os.neva
emil14: Addressed in b1111d0b.

The `std/os` API intentionally mirrors Go `os` names where these components are thin extern wrappers. The Neva-specific adaptation is in the result shape: Go multi-return/interface data is represented as explicit typed values such as `LookupEnvResult`, `DirEntry`, and `FileInfo`, because Neva has typed message values and ports rather than Go tuples/interfaces.

I added comments to those public result types to make that relationship explicit.

e2e/os_fs/main/main.neva
emil14: Addressed in b1111d0b.

I removed explicit `builtin` imports/usages from the OS e2e files touched by this PR and applied port omission where it is unambiguous. I also created https://github.com/nevalang/neva/issues/1142 for the compiler-level rule to forbid explicit `builtin` imports, since that is broader than this PR.

I did not deeply decompose `os_fs` in this commit: after removing the noisy `builtin` qualification and redundant port names, it is still a single end-to-end filesystem orchestration, so splitting it further would be a separate readability refactor rather than a targeted fix for the current review comments.

e2e/os_process/main/main.neva
emil14: Addressed in b1111d0b.

`Getppid` is the Go `os.Getppid` wrapper: it returns the parent process identifier. I renamed the e2e node from `getppid` to `get_parent_pid` and clarified the public std/os comment so the meaning is visible at the Neva API level.

internal/runtime/funcs/args_test.go
emil14: Addressed in b1111d0b.

I expanded runtime unit coverage beyond the existing benchmark/e2e layer:

- args message construction now covers empty, single, and multi-arg cases.
- env runtime funcs cover getenv, lookup_env missing result, setenv success/error, unsetenv, expand_env, and clearenv.
- process runtime funcs cover getwd, chdir success/error, getpid, getppid, hostname, and executable.
- filesystem runtime funcs cover mode conversion, dir/file info conversion, mkdir/mkdir_all/read_dir/remove/remove_all/rename/stat/lstat/truncate/temp_dir/mkdir_temp/create_temp.

I also added the durable rule to the Go authoring skill: runtime function changes need focused unit tests, not only benchmarks/e2e coverage.

internal/runtime/funcs/utils_os.go
emil14: Done in b1111d0b: renamed `os_helpers.go` to `utils_os.go` and renamed the matching test file to `utils_os_test.go`.

internal/runtime/funcs/utils_os.go
emil14: Done in b1111d0b: added doc comments for every helper function in `utils_os.go`.

internal/runtime/funcs/utils_os.go
emil14: I kept these two small helpers in b1111d0b, but documented them.

The saving is not line count; it is keeping the uniform setup error wrapping in one place. These helpers are used by the loop creators so the creators stay focused on wiring the runtime loop shape, while the low-level `io.In.Single` / `io.Out.Single` error context is consistent everywhere.

If we later decide the indirection is not worth it, inlining is mechanically easy, but I think the current shape is clearer for this shared std/os runtime support.

internal/runtime/funcs/utils_os.go
emil14: I kept the check in the error path in b1111d0b.

`errOut == nil` is only possible for helpers whose step is expected not to fail. If such a step returns an error anyway, that is an internal invariant violation, so panic is the right behavior. Checking inside the error branch means the normal path pays no extra cost and the same loop helper can support both no-error and error-returning externs.

We could assert at setup time only by splitting the loop helpers or requiring an err outport even for funcs that cannot fail; that would either duplicate the loop logic or make the runtime API noisier. The current check is narrow and only runs when an unexpected error already happened.

Issue #908, opened 2025-10-11
Redesign of `stream`


emil14:
# Current Design

```neva
pub type stream<T> struct {
    data T
    idx int
    last bool
}
```

- every item has data, no real "bracket" messages
- index is zero? it's first item
- last is true? it's last item

# Proposed Changes

1. Use tagged unions instead of struct 
2. Get rid of `idx` to get rid of problems related to indexing (and make messages more light-weight)

## Tagged Unions instead of Struct

```neva
pub type stream<T> union {
    Open
    Data StreamData<T>
    Close
}

pub type StreamData<T> struct {
    idx int
    data T
}
```

- first and last items in stream are FBPs "brackets" and do not carry data
- thus components that care about state of the stream item must use pattern matching
- instead of asking if item is last we ask if stream is closed

## Get rid of `idx`

This way we don't need `StreamData` type and `stream` becomes just this:

```neva
pub type stream<T> union {
    Open
    Data T
    Close
}
```

### Upsides

- no issues with maintaining the correct order of the indexes (example: #763)
- no extra 64bits of memory for index inside every stream item

### Downsides

- stream processors (e.g. mapper or filterer) do not know about the index of the current element, which is not the API of map/filter/reduce in JS (but we can check other implementations in functional languages)

### Notes

- It's possible to have a higher-order wrapper like `IndexedStreamItem<T>` that wraps normal stream item and also carries the index, maybe it's possible to only use it when needed, one must perform research and brainstorm this

PR #1042, opened 2026-02-24
Redesign stream as union and add streams.Enumerate (#908)

emil14:
## Context
Implements core migration for #908 in scope of tracker #1033.

## What is done
- migrate `stream<T>` from struct to tagged union (`Open | Data T | Close`)
- add parser/analyzer support for generic union literals in sender positions
- migrate runtime stream funcs to Open/Data/Close protocol
- add `streams.Indexed<T>`, `streams.Enumerate<T>`, `streams.Just<T>`
- add runtime stream union helpers (`IsData`/`IsClose`/`UnwrapData`)
- stabilize `streams.ForEach` via runtime controller (`stream_for_each_controller`) to preserve ordering and backpressure
- fix `streams.Filter` event routing under union-stream semantics
- remove implicit non-union->Data coercion in stream helper path

## Verification
Passing:
- `go test ./internal/runtime/funcs ./examples/filter_list -count=1 -timeout 5m`
- `go test ./examples/for_loop_over_list -count=1 -timeout 8m`
- `go test ./examples/filter_list ./examples/for_loop_over_list -count=1 -timeout 9m`

Note:
- local `golangci-lint` binary in this environment is built with older Go (`go1.25`), so direct `golangci-lint run ./...` is blocked against target `go1.26.0`.

Refs #908
Refs #1033


std/streams/streams.neva
```diff
@@ -27,16 +35,32 @@ pub def FromDict<T>(data dict<T>) (res stream<DictEntry<T>>)
 #extern(stream_int_range)
 pub def Range(from int, to int) (res stream<int>)
 
+// Enumerate attaches a 0-based index to each Data item in a stream.
+#extern(stream_enumerate)
+pub def Enumerate<T>(data stream<T>) (res stream<Indexed<T>>)
+
+// IsData returns true when stream item has Data tag.
+#extern(stream_is_data)
+def IsData<T>(data stream<T>) (res bool)
```
emil14: are you sure we should have these isdata/close components (also i dont see isopen), given the fact that stream element is a union and we can pattern-match on the tag? also if these components would be needed I would rather implement them as neva components using switch/match instead

std/streams/streams.neva
```diff
@@ -27,16 +35,32 @@ pub def FromDict<T>(data dict<T>) (res stream<DictEntry<T>>)
...
+def UnwrapData<T>(data stream<T>) (res T)
+
 // --- Synchronization ---
 
-// Wait blocks until last stream item arrive, then sends a signal.
+// Wait blocks until stream close item arrives, then sends a signal.
 pub def Wait(data stream<any>) (sig any) {
     del Del
-    s Switch<bool>
+    s Switch<stream<any>>
```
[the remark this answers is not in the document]
emil14: Renamed the node in `Wait` to `switch` to match the style guide.

std/streams/streams.neva
```diff
@@ -1,5 +1,10 @@
 // --- Sources ---
 
+pub type Indexed<T> struct {
```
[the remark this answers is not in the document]
emil14: Renamed the type to `Enumerated<T>` and updated `Enumerate` plus the emitted struct shape to use `idx` and `item`.

emil14 (commented):
# Review Summary

## Critical Issues (Blocking)

- `internal/compiler/parser/listener_helpers.go:798` now gives every union literal sender a tag-only type (`map[tag]nil`), and `internal/compiler/typesystem/subtype_checker.go:152` then allows that tag-only expression to satisfy a typed union member. That makes payload-bearing literal senders unsound: a program like `type U union { Int int }` plus `:start -> U::Int('bad') -> Pass<U> -> :stop` compiles and runs successfully, even though the `Int` payload is a string. Please preserve/check the wrapped payload type for value literals, or separate pattern-only literals used for switch cases from value literals used as actual messages, and add a regression test that rejects this mismatch.

## Non-critical Issues

- `internal/runtime/funcs/range_int.go:29`, `internal/runtime/funcs/string_join.go:30`, and `internal/runtime/funcs/string_join.go:136` changed multi-input components from concurrent receives to ordered receives. These components previously used `receive2`, and the repo-local Go guidance calls out this class of runtime deadlock risk for multi-input helpers. Please restore concurrent receive behavior, or add an explicit test proving that reversed arrival order cannot block these components. While touching `Range`, also avoid `rangeValues` materializing the whole range before sending; `streams.Range` should stream `Data` events directly instead of allocating an O(n) slice for large ranges.

## Conclusion

Do not merge until the union literal payload type hole is fixed. GitHub also reports this PR as `CONFLICTING` against `main`, so it needs a rebase/merge update before it can land.

e2e/type_expr_with_imported_type_arg/main/main.neva
```diff
@@ -3,15 +3,13 @@
 // can be resolved by type-system, because there was a bug with it
 
 import { @:foo }
+import { streams }
```
emil14: follow style guide and how single import statement per file, fetch fresh main and check if theres no analyzer check for that and add if none, with unit tests too

e2e/type_expr_with_imported_type_arg/main/main.neva
```diff
@@ -3,15 +3,13 @@
...
 	---
-	:start -> [
-		0 -> [justFoo:bar, itemWithFoo:idx],
-		true -> itemWithFoo:last
-	]
-	justFoo -> itemWithFoo:data
-	itemWithFoo -> :stop
+	:start -> 0 -> justFoo:bar
+	justFoo -> just:data
+	just:res -> :stop
```
emil14: :start -> 0 -> justFoo -> just -> :stop 

follow style guide and use chained connections whenever possible

e2e/type_expr_with_imported_type_arg/main/main.neva
```diff
@@ -3,15 +3,13 @@
 // can be resolved by type-system, because there was a bug with it
 
 import { @:foo }
+import { streams }
 
 def Main(start any) (stop any) {
 	justFoo Struct<foo.Foo>
-	itemWithFoo Struct<stream<foo.Foo>>
+	just streams.Item<foo.Foo>
```
emil14: justFoo => foo_struct

just => foo_stream_item

e2e/union_literal_payload_type_mismatch/main/main.neva
```diff
@@ -0,0 +1,10 @@
+type U union {
+	Int int
+}
+
+def Main(start any) (stop any) {
+	pass Pass<U>
+	---
+	:start -> U::Int('bad') -> pass
+	pass -> :stop
```
emil14: chain

examples/image_png/main.neva
```diff
@@ -1,4 +1,4 @@
-import { io, image, fmt, runtime }
+import { io, image, fmt, runtime, streams }
```
emil14: lets actually forbit `,` syntax for `import{}` on syntax level (grammar, antlr)

the import must always be multi-line

```neva
import {
  io
}
```

even if theres only one package imported

this allows not to think which variatn to use

this would take updating g4 grammar, regen antlr, perhaps updating the listener helpers (maybe not), almost certainly the smoke and perhaps unit tests in parser and a TON of *.neva code in examples and e2e

examples/image_png/main.neva
```diff
@@ -43,8 +32,8 @@ def Main(start any) (stop any) {
 		'minimal.png' -> writeAll:filename
 	]
 	newColor -> newPixel:c
-	newPixel -> newStream:p
-	newStream:s -> new
+	newPixel -> item:data
+	item:res -> new
```
emil14: chain!

internal/compiler/analyzer/network.go
```diff
@@ -868,6 +868,14 @@ func (a Analyzer) getConstSenderType(
 		}
 	}
 
+	if constSender.Value.Message.Union != nil {
```
emil14: if you change something, reflect that in unit tests

internal/compiler/analyzer/union_logic.go
```diff
@@ -6,13 +6,120 @@ import (
...
 	src "github.com/nevalang/neva/pkg/ast"
+	"github.com/nevalang/neva/pkg/core"
 )
 
 type unionActiveTagInfo struct {
 	tag         string
 	tagTypeExpr ts.Expr
 }
 
+func (a Analyzer) unionLiteralSenderType(
```
emil14: I cant see what this function does by its name, rename it as a verb and make it clear

internal/compiler/analyzer/union_logic.go
```diff
@@ -6,13 +6,120 @@ import (
 	"github.com/nevalang/neva/internal/compiler"
 	ts "github.com/nevalang/neva/internal/compiler/typesystem"
 	src "github.com/nevalang/neva/pkg/ast"
+	"github.com/nevalang/neva/pkg/core"
```
emil14: why do we have analyzer changes in the first place? wasnt this PR about changing stdlib and runtime functions?

just answer, lets talk, I wanna understand

internal/compiler/typesystem/subtype_checker.go
```diff
@@ -57,49 +59,64 @@ func (s SubtypeChecker) Check(
...
-
-	return s.checkLitSubtype(&expr, &constr, &params)
-}
-
-func (s SubtypeChecker) checkInstSubtype(expr, constr *Expr, params *TerminatorParams) error {
-	isSubTypeRecursive, err := s.terminator.ShouldTerminate(params.SubtypeTrace, params.Scope)
-	if err != nil {
-		return fmt.Errorf("%w: %w", ErrTerminator, err)
-	}
+		isSubTypeRecursive, err := s.terminator.ShouldTerminate(
```
emil14: same question as for the analyzer, why we changing something in the ts?

std/streams/filter.neva
```diff
@@ -1,29 +1,29 @@
...
 pub def Filter<T>(data stream<T>) (res stream<T>) {
     split Split<T>{predicate IPredicate<T>}
-    s2l lists.FromStream<T>
-    l2s FromList<T>
+    del Del
     ---
     :data -> split
-    split:then -> s2l -> l2s -> :res
+    split:then -> :res
+    split:else -> del
```
emil14: we can just omit split:else and not use Del explicitly cant we?

std/streams/filter.neva
```diff
@@ -1,29 +1,29 @@
...
 }
 
 // Split splits a stream into two streams based on a predicate.
-// Notice that it does not modifies stream item indexes and last flag.
 pub def Split<T>(data stream<T>) (then stream<T>, else stream<T>) {
-    cond Cond<stream<T>>
+    controller SplitController<T>
```
emil14: lets use ctrl for such long word

std/streams/filter.neva
```diff
@@ -1,29 +1,29 @@
...
+    controller:then -> :then
+    controller:else -> :else
 }
 
 // IPredicate represents a logic that determines if a value satisfies a condition.
 // It is used by components like Filter and Split to decide how to process stream elements.
 pub interface IPredicate<T>(T) (bool)
+
+#extern(stream_split_controller)
+def SplitController<T>(data stream<T>, pass bool) (item T, then stream<T>, else stream<T>)
```
emil14: Split controller must have doc comments and tis name is not descriptive and a bit long

std/streams/filter.neva
```diff
@@ -1,29 +1,29 @@
...
 // Split splits a stream into two streams based on a predicate.
-// Notice that it does not modifies stream item indexes and last flag.
 pub def Split<T>(data stream<T>) (then stream<T>, else stream<T>) {
-    cond Cond<stream<T>>
+    controller SplitController<T>
     predicate IPredicate<T>
     ---
-    :data -> [cond:data, .data -> predicate]
-    predicate -> cond:if
-    cond:then -> :then
```
emil14, line 23: I personally liked that split was implemented just using cond, now it has dedicated runtime func. Is this justified?

internal/runtime/funcs/utils.go
```diff
@@ -351,3 +344,52 @@ func traceTreeFromHop(tracer *runtime.Tracer, hop runtime.TraceHop, visited map[
 	delete(visited, hop.Index)
 	return tree, true
 }
+
+const (
+	streamTagOpen  = "Open"
```
emil14: if possible, lets use uint8 for that, you can implement String() (stringer) if needed for debugging

internal/runtime/funcs/utils.go
```diff
@@ -351,3 +344,52 @@ func traceTreeFromHop(tracer *runtime.Tracer, hop runtime.TraceHop, visited map[
...
+	streamTagData  = "Data"
+	streamTagClose = "Close"
+)
+
+func streamOpen() runtime.UnionMsg {
```
emil14: move all streams related utils to `utils_streams.go`

internal/runtime/funcs/utils.go
```diff
@@ -351,3 +344,52 @@ func traceTreeFromHop(tracer *runtime.Tracer, hop runtime.TraceHop, visited map[
...
+func streamData(data runtime.Msg) runtime.UnionMsg {
+	return runtime.NewUnionMsg(streamTagData, data)
+}
+
+func streamClose() runtime.UnionMsg {
+	return runtime.NewUnionMsg(streamTagClose, nil)
+}
+
+func streamUnion(msg runtime.Msg) runtime.UnionMsg {
+	defer func() {
```
emil14: what is this fucking abomination in the runtime? remove defer, and this helpper is useless overall

if you 100% SURE that you absolutely need this, you can assert that tag is one of enums uint8 values

internal/runtime/funcs/utils.go
```diff
@@ -351,3 +344,52 @@ func traceTreeFromHop(tracer *runtime.Tracer, hop runtime.TraceHop, visited map[
 	delete(visited, hop.Index)
 	return tree, true
 }
+
+const (
```
emil14: I feel like stream item is a special case of Union that runtime can be aware of. I.e. moving this to `runtime/stream.go`

Or maybe to `msg.go` (or message.go it is?) in the runtime under the `Streams` section.

Like runtime functions are implementation of stuff, not the layer that dictates the structure

Same perhaps goes for `error` though

std/streams/for_each.neva
```diff
@@ -1,47 +1,20 @@
-// ForEach applies a handler to each item in a stream to produce side-effects.
-// It never modifies the stream.
-//
```
emil14, line 3: you removed a lot of useful comments, arent they still valid?

std/streams/for_each.neva
```diff
@@ -1,47 +1,20 @@
...
-    // IDEA - if we use sync.Step, we don't have to mess with indexes
-
-    first IsFirst<T> // to check if current msg is the first
-    lock1 Lock<stream<T>> // to lock current msg before sending to handler
-    lock2 Lock<stream<T>> // to lock current msg before sending downstream
-    handler ISideEffector<T>? // to process current msg
-    pass1 Pass<any> // to unlock current msg to send to handler
-    pass2 Pass<any> // to unlock current msg to send downstream
+    controller ForEachController<T>
+    handler ISideEffector<T>
```
emil14: lets stick with naming `IForEachHandler` here

std/streams/for_each.neva
```diff
@@ -1,47 +1,20 @@
...
 pub def ForEach<T>(data stream<T>) (res stream<T>, err error) {
-    // IDEA - if we use sync.Step, we don't have to mess with indexes
-
-    first IsFirst<T> // to check if current msg is the first
-    lock1 Lock<stream<T>> // to lock current msg before sending to handler
-    lock2 Lock<stream<T>> // to lock current msg before sending downstream
-    handler ISideEffector<T>? // to process current msg
-    pass1 Pass<any> // to unlock current msg to send to handler
-    pass2 Pass<any> // to unlock current msg to send downstream
+    controller ForEachController<T>
```
emil14: Im not sure if I love this "controller" pattern I see here and there... we need to talk about this

std/streams/map.neva
```diff
@@ -1,19 +1,17 @@
 // IMapHandler is a dependency for Map.
 pub interface IMapHandler<T, Y>(T) (Y)
 
-// Map maps one stream onto another.
-// Unlike ForEach, it does not block the stream. As soon as one item processed,
-// it sends an item downstream and processes the next one.
-// It's can receive next message before previous one is processed,
```
emil14, line 7: same about comments, controller

std/streams/reduce.neva
```diff
@@ -6,17 +6,27 @@ pub interface IReducer<T, Y>(left T, right T) (res Y)
 // Reduce applies a reduction component to a stream of messages, accumulating the result.
 // It takes an initial value and a stream of data, and produces a single result.
 pub def Reduce<T, Y>(data stream<T>, init Y) (res Y) {
+    del_open Del
+    del_else Del
```
emil14: remember we can delete msgs explicitly by not using the outports, maybe this must be (if not yet) part of the stykle guide

std/streams/reduce.neva
```diff
@@ -6,17 +6,27 @@ pub interface IReducer<T, Y>(left T, right T) (res Y)
...
     reducer IReducer<T, Y>
     acc Accumulate<Y>
+    sw Switch<stream<T>>
```
emil14: sw -> switch

std/streams/reduce.neva
```diff
@@ -6,17 +6,27 @@ pub interface IReducer<T, Y>(left T, right T) (res Y)
...
+    reducer -> acc:upd
 }
 
 // Accumulate maintains the current state of the reduction.
-// It updates its value with each new input and outputs the final result when last is true.
+// It updates its value with each new input and outputs the current result on flush.
 #extern(accumulator)
-def Accumulate<T>(init T, upd T, last bool) (cur T, res T)
\ No newline at end of file
+def Accumulate<T>(init T, upd T, flush any) (cur T, res T)
```
emil14, line 30: why did you decide to change the acc API?

std/streams/streams.neva
```diff
@@ -15,6 +20,9 @@ pub def FromList<T>(data list<T>) (res stream<T>)
 // must remain whole stream elements.
 #extern(string_to_stream)
 pub def FromString(data string) (res stream<string>)
+// Item creates a single-item stream.
```
emil14: use Neva's parsable doc-comments format for all Neva entities you're touching in this PR at least

this specific component lacks explanation why we need it

for me now for example its not obvilos how it helps if the stream always consist of one item

why do we need such streams? what for?

std/streams/streams.neva
```diff
@@ -27,19 +35,23 @@ pub def FromDict<T>(data dict<T>) (res stream<DictEntry<T>>)
...
     :data -> [
-        .last -> s:data,
-        true -> s:case[0]
+        switch:data,
+        stream<any>::Close -> switch:case[0]
     ]
-    s:case[0] -> :sig
-    s:else -> del
+    switch:case[0] -> :sig
+    switch:else -> del
```
emil14: how about omitting else

std/streams/streams.neva
```diff
@@ -1,5 +1,10 @@
 // --- Sources ---
 
+pub type Enumerated<T> struct {
```
emil14, line 3: lets talk about this

how do I now its last el of the stream? its closed right? and I always work with stream<Enumirated<T>> and not just Enumirated<T>?

how do I know how many items are there? thats a better question right?

I guess sometimes we know and sometimes we dont so `len` can be either `uint64` with possible zero value - since there are no empty streams by definition (right? or no?) we can assign `0` to `unknown`

OR it might be `maybe<uint64>` which is another option

or is this all too much and bad for perf?

e2e/type_expr_with_imported_type_arg/main/main.neva
emil14: Addressed in 0e48e7d9.

I converted this to a single multiline import block and applied the same syntax across repo-local `.neva` examples/e2e/benchmarks. I also updated the parser grammar, regenerated ANTLR artifacts, documented the rule in `docs/style_guide.md`, and added parser tests rejecting single-line and comma-separated import blocks.

e2e/type_expr_with_imported_type_arg/main/main.neva
emil14: Addressed in 0e48e7d9.

I chained the stream item part as `foo_struct -> foo_stream_item -> :stop`. I kept `:start -> 0 -> foo_struct:bar` separate because `Struct<foo.Foo>` has the named input `bar`; omitting that port makes desugaring fail since there is no inferable first inport in that generated struct builder shape.

e2e/type_expr_with_imported_type_arg/main/main.neva
emil14: Done in 0e48e7d9: renamed the nodes to `foo_struct` and `foo_stream_item`.

e2e/union_literal_payload_type_mismatch/main/main.neva
emil14: Done in 0e48e7d9: collapsed this to `:start -> U::Int('bad') -> pass -> :stop`.

examples/image_png/main.neva
emil14: Implemented in 0e48e7d9 for repo-local code.

Changes:
- grammar now requires import block items to be multiline and removed comma support from import definitions;
- ANTLR artifacts regenerated;
- parser regression tests added for `import { fmt }` and comma-separated imports;
- repo-local `.neva` files converted to multiline import blocks;
- style guide updated with the new rule.

One important caveat: full `go test ./e2e/...` now exposes that external dependency `github.com/nevalang/x@0.0.17` still uses `import { fmt }`, and even its current `main` still has that syntax. So local enforcement is done, but the ecosystem dependency needs a migration/tag before the external-import e2e can pass under the new grammar.

examples/image_png/main.neva
emil14: Done in 0e48e7d9: changed the image example to `newPixel -> item -> new`.

internal/compiler/analyzer/network.go
emil14: Acknowledged. The analyzer call-site change in this commit is only a rename to the clearer helper name, so there is no analyzer behavior change to add a new analyzer unit for.

For functional changes in this batch, I added parser tests for the import syntax rule and runtime tests for the new stream runtime API. Existing analyzer/e2e coverage still covers the union literal payload behavior touched by this PR.

internal/compiler/analyzer/union_logic.go
emil14: Done in 0e48e7d9: renamed `unionLiteralSenderType` to `inferUnionLiteralSenderType`, which makes it a verb phrase and describes that it infers the sender type from a union literal.

internal/compiler/analyzer/union_logic.go
emil14: The analyzer changes are needed because once stream items became tagged unions, union literals started being used as actual senders, not just type declarations. The analyzer has to infer the precise literal sender type and validate payload compatibility, otherwise cases like `U::Int('bad')` can slip too far through the pipeline or produce worse errors.

So the PR is mostly stdlib/runtime surface, but the compiler has to understand the new union-literal sender shape to keep type errors precise.

internal/compiler/typesystem/subtype_checker.go
emil14: The typesystem change is for subtype checks involving literal union expressions. Stream events are represented as tagged union values, and the analyzer needs `Data<T>`-style literal union sender types to be checked against the declared union type. Without the subtype checker understanding that narrower union literals can fit broader union constraints, the stream-union representation becomes awkward or rejects valid code.

std/streams/filter.neva
emil14: Done in 0e48e7d9: removed the explicit `Del` in `Filter` and omitted the unused `split:else` flow.

std/streams/filter.neva
emil14: Done in 0e48e7d9: renamed the node to `ctrl`.

std/streams/filter.neva
emil14: Done in 0e48e7d9: renamed the private extern component to `SplitCtrl` and added parsable doc comments with inport/outport tags.

std/streams/filter.neva
emil14: This is justified by the new stream representation. With stream events as `Open/Data/Close`, `Split` has to preserve event ordering while waiting for a predicate result for each Data item and forwarding Open/Close to both output streams. A plain `Cond` can route a single value, but it cannot coordinate that stream protocol and backpressure by itself without extra glue. The runtime controller is that glue.

internal/runtime/funcs/utils.go
emil14: Done in 0e48e7d9: stream tags are now represented by `runtime.StreamTag uint8`, with `String()` returning the Neva union tag name.

internal/runtime/funcs/utils.go
emil14: I went one layer lower in 0e48e7d9: the stream helpers are no longer in `internal/runtime/funcs/utils.go` at all. They now live in `internal/runtime/stream.go`, because the stream message shape is a runtime-level concept rather than a funcs-package utility.

internal/runtime/funcs/utils.go
emil14: Fixed in 0e48e7d9 by removing the funcs-level helper entirely. The new runtime API uses `runtime.AsUnion` and explicit tag checks, so there is no `defer`/`recover` conversion path anymore.

internal/runtime/funcs/utils.go
emil14: Agreed and fixed in 0e48e7d9. Stream message constructors/predicates/data extraction now live in `internal/runtime/stream.go`, and runtime funcs call that API instead of dictating stream structure from `internal/runtime/funcs`.

std/streams/for_each.neva
emil14: Addressed in 0e48e7d9: restored the useful ForEach comments and adapted them to the new stream-union design/backpressure behavior.

std/streams/for_each.neva
emil14: Done in 0e48e7d9: renamed the interface back to `IForEachHandler`.

std/streams/for_each.neva
emil14: Agreed that this needs design attention. My current reasoning is that these controllers are runtime protocol adapters: they preserve `Open/Data/Close` ordering and introduce per-item backpressure around user handlers. That is hard to express cleanly with today’s generic graph primitives.

I did not rename/remove the whole pattern in this commit beyond the targeted fixes, because doing that well probably needs a broader design pass over Map/ForEach/Split together.

std/streams/map.neva
emil14: Addressed partly in 0e48e7d9: restored/adapted Map comments and added parsable doc comments for `MapController`.

I kept the controller shape for the same reason as ForEach/Split: it is coordinating stream protocol ordering and backpressure around handler output. I agree the pattern deserves a broader follow-up discussion.

std/streams/reduce.neva
emil14: Applied where it was directly relevant in 0e48e7d9: `Filter` and `Wait`/`Reduce` no longer route unused branches into explicit `Del` nodes where the outport can simply be omitted.

I did not add a separate style-guide rule for explicit deletion in this commit, beyond the existing guidance to omit ports when possible, because that deserves wording that is not too broad for cases where explicit synchronization/deletion is intentional.

std/streams/reduce.neva
emil14: Done in 0e48e7d9: renamed `sw` to `switch`.

std/streams/reduce.neva
emil14: The accumulator API changed because stream items no longer carry a `last` boolean. End-of-stream is now represented by the `Close` union event, so Reduce needs to flush on Close rather than on a `last` field inside every item.

That keeps the stream item payload clean and moves stream lifecycle information into the stream protocol itself (`Open/Data/Close`).

std/streams/streams.neva
emil14: Addressed in 0e48e7d9: added parsable doc comments for `Item`, including why a single-item stream is useful and the inport/outport semantics.

std/streams/streams.neva
emil14: Done in 0e48e7d9: omitted the unused `else` branch in `Wait` and removed the explicit `Del`.

std/streams/streams.neva
emil14: I did not add length metadata in this commit. With the current design, `Enumerated<T>` is intentionally just `{ idx, item }`, and stream completion is represented by the `Close` event around the stream, not by the item itself.

For total length, I think `maybe<uint64>` is semantically cleaner than reserving `0` for unknown, because empty streams may become possible or useful later and `0` as sentinel would leak protocol assumptions into user data. But adding length would also require sources/combinators to preserve or deliberately drop that metadata, which is a bigger design question than Enumerate itself.

std/streams/streams.neva
```diff
@@ -15,6 +20,9 @@ pub def FromList<T>(data list<T>) (res stream<T>)
 // must remain whole stream elements.
 #extern(string_to_stream)
 pub def FromString(data string) (res stream<string>)
+// Just creates a single-item stream.
+#extern(stream_just)
+pub def Just<T>(data T) (res stream<T>)
```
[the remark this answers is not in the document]
emil14: Resolved by the current API: the constructor is now `streams.Item<T>`, not `Just`. That matches the stream vocabulary better and avoids importing FP naming into the public stdlib API.

std/streams/streams.neva
emil14: Resolved in the current shape: the separate `isdata`/`close` components are not part of the public `std/streams` API. Stream events are matched as union tags (`Open`, `Data`, `Close`) with normal `Switch`/matching patterns.

std/streams/streams.neva
```diff
@@ -27,16 +35,32 @@ pub def FromDict<T>(data dict<T>) (res stream<DictEntry<T>>)
...
+#extern(stream_is_data)
+def IsData<T>(data stream<T>) (res bool)
+
+// IsClose returns true when stream item has Close tag.
+#extern(stream_is_close)
+def IsClose<T>(data stream<T>) (res bool)
+
+// UnwrapData extracts payload from stream<T>::Data item.
+#extern(stream_unwrap_data)
+def UnwrapData<T>(data stream<T>) (res T)
```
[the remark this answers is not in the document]
emil14: Agreed; the current `Wait` implementation uses `Switch<stream<any>>` to match `Close` rather than a dedicated public component for this case.

examples/image_png/main.neva
emil14: Update ALL the neva code in this repo so its valid (except it needs to be not valid for testing purposes) AND create PR for the `x` repo with the proper fix, and as soon as that PR merged (and this one too) we will release new version of the X (and then of Neva)

emil14: Updated the Neva code in this repo after merging current `main`; the only remaining `import { ... }` occurrences are an intentionally invalid parser fixture and an old syntax example inside a comment. I also opened the companion `nevalang/x` PR for the package-side import syntax fix: https://github.com/nevalang/x/pull/12. Once that PR and this one are merged, `x` can be released with the compatible source.

benchmarks/atomic/streams_map/main.neva
```diff
@@ -1,8 +1,12 @@
 // Benchmarks the atomic streams.Map hot path with a neutral pass-through handler.
 // FromList and FromStream are kept only as the minimal support wiring for a
 // valid standalone stream program.
-import { lists }
-import { streams }
+import {
+	lists
+}
+import {
```
emil14, line 7: having two import statements must be forbidden by analyzer (with unit tests)

emil14: Done in 7f2839b4. The parser now preserves import-statement block metadata in the AST, and the analyzer rejects files with more than one import statement. I added unit coverage on both sides: parser verifies that multiple blocks are tracked, and analyzer verifies that the second block is rejected with a file-level error. I also bumped the repo manifests from `github.com/nevalang/x` 0.0.17 to the newly released 0.0.18.

internal/compiler/analyzer/union_logic.go
emil14: You are right to challenge this, and my previous wording was too broad.

Union literals as network senders are not new in this PR. They were introduced with tagged unions in [v0.32.0](https://github.com/nevalang/neva/releases/tag/v0.32.0), whose release notes explicitly show both tag-only and payload sender forms, for example `Day::Monday -> println` and `Input::Int(42) -> println`. Later, [v0.34.0](https://github.com/nevalang/neva/releases/tag/v0.34.0) removed the old composite/wrapping union sender form (`Type::Tag(data) -> handler`) and required explicit `Union<T>` wiring instead. But tag-only union literals still remain part of the chain-only literal/const-sender model, for example the v0.34 migration shows `:start -> MyUnion::Tag -> union:tag`.

So the reason for compiler changes here is **not** “this PR introduces union senders”. It does not.

The stream-specific pressure is narrower: this PR changes builtin `stream<T>` from a struct shape into a generic tagged union:

```neva
pub type stream<T> union {
    Open
    Data T
    Close
}
```

After that, stdlib stream components use generic tag-only union literals as `Switch` cases, for example:

```neva
stream<T>::Close -> switch:case[1]
stream<T>::Data -> switch:case[2]
```

`Data` is a typed tag (`Data T`), but in this position it is intentionally used as a pattern/case literal, not as construction of a concrete `stream<T>::Data(payload)` value. The old analyzer/typesystem path was not precise enough for this generic tag-literal-as-pattern case, especially around resolving `stream<T>` before concrete type-argument substitution and allowing a tag-only case literal to match a typed union member.

There is a separate broader tightening in this PR around invalid payload construction, covered by the `U::Int(bad)` regression test. That part is not a new stream feature either; it is an existing union-literal validation gap that became visible while exercising the stream-union representation. So the accurate framing is: the PR mostly changes stdlib/runtime streams, but it also needs a small compiler fix for generic union tag literals used by `Switch`, plus a related regression test for existing union literal payload validation.

I agree the compiler diff deserves this stricter framing. If we want to be even more conservative, I can also look at narrowing the `subtype_checker.go` diff so it contains only the semantic change needed for tag-only union case matching, without unrelated refactor noise.

internal/compiler/typesystem/subtype_checker.go
emil14: I rechecked this from first principles and need to correct the earlier answer in this thread. The broad subtype change is not justified merely by changing `stream<T>` to a union.

There are two separate concerns:

1. Generic union-literal syntax. The stdlib now wants to write `stream<T>::Open`, `stream<T>::Data`, and `stream<T>::Close` in generic components. The pre-PR grammar only accepted `entityRef :: tag`, so it could not express the explicit `stream<T>` instantiation. Supporting that spelling requires the grammar/AST/parser work.

2. A pre-existing soundness issue for tag-only literals of typed union members. This is independent of generics and is reproducible on current `main`:

```neva
type U union {
    Data int
}

def Main(start any) (stop any) {
    pass Pass<U>
    ---
    :start -> U::Data -> pass -> :stop
}
```

This compiles and runs even though `U::Data` carries no `int` payload. More seriously, if that same tag-only `U::Data` is sent to `Switch<U>:data`, while `U::Data` is also used as the case pattern, the analyzer infers `int` for the case output. At runtime the switch cannot unbox a nil payload, forwards the union message itself, and an `Add` consuming the supposedly-`int` case output panics with `unexpected Int method call on internal message type`.

Before this PR, the compiler avoided the immediate generic-argument error for non-generic unions by assigning a union literal the nominal type `U`, then checking `U <: U`. That is why the problem was latent: it did not model the difference between a complete `Data(value)` and a tag-only `Data` pattern. Generic `stream<T>` exposed the missing syntax, but did not create this underlying issue.

The current PR instead represents a tag-only literal structurally as `{ Data: nil }` and adds the generic rule `{ Data: nil } <: { Data: T }`. That makes the stream case compile, but it still permits the tag-only pattern to travel through ordinary `stream<T>` ports, so it does not solve the runtime/type mismatch above. It also makes the interpretation global rather than a narrowly defined language rule.

We need to decide the language semantics explicitly: either a tag-only literal for a typed member is only valid in pattern/tag positions, or it becomes a distinct first-class tag/pattern concept. The latter is a broader language-design task. For this PR, the immediate compiler change should be narrowed and tested deliberately; the general typed-tag/payload soundness issue needs a linked high-priority issue rather than being silently encoded as a broad subtype relaxation. I will not resolve this discussion until that boundary and the intended PR scope are agreed.

emil14: Implemented and pushed in 3ed17a16 (follow-up lint annotations: 1079880e).

Final rule in this PR:
- A tag-only literal such as `Event<T>::Data` is a pattern, not a complete `Event<T>` value. The generic subtype checker again rejects `{Data:nil} <: {Data:T}`.
- The analyzer has the only two contextual exceptions: `Switch:case` and `Union:tag`. There it validates that the receiver is a union and contains the selected tag; runtime matching uses the tag only, so no payload is claimed or unboxed.
- Ordinary ports keep normal subtype validation. Consequently `Event::Data -> Pass<Event>` is rejected instead of compiling into a later nil-payload failure.
- A generic union pattern must be explicit: `Event<T>::Data`. Bare `Event::Data` now fails rather than falling back to the raw generic definition. A narrowly scoped fallback remains only to inspect member names when the analyzer stage no longer has the formal `T` in scope.

Regression coverage added:
- `e2e/union_tag_cannot_be_value`: rejects a typed tag-only literal passed as a union value.
- `e2e/generic_union_tag_requires_type_args`: rejects a bare generic union tag.
- Existing union/Switch tests plus stream `Reduce` e2e pass.

This fixes the concrete compiler soundness path needed by the stream redesign. A first-class general tag/pattern type remains a separate language-design question; it is not introduced here.

emil14: Correction to my previous reply: I reverted the implementation it described. It was wrong to add compiler-aware exceptions that allow a tag-only literal only at Switch:case and Union:tag, or to introduce a new policy for bare generic tags in this PR.

The conclusion from this discussion is instead:
- Changing stream<T> to a tagged union does not by itself justify a context-sensitive subtype exception.
- Switch and Union must not become semantic special cases recognized by the compiler. Any valid solution has to be expressed through a general language/type-system mechanism usable by arbitrary components.
- The existing global rule that lets a tag-only union literal match a typed member remains an unresolved type-system question. It must not be treated as accepted by this PR without a coherent model for a union member used as a tag/pattern.
- The meaning and validity of Event::Data for a generic Event<T> belongs to that same design work; this PR adds no rule for it.

The regression tests and analyzer logic from my earlier reply were reverted because they encoded the wrong contextual rule. The compiler/type-system problem should be tracked and designed separately as a high-priority follow-up, rather than papered over here.

internal/compiler/analyzer/union_logic.go
emil14: Split the general union-literal payload inference out of #1042. Commit 10839ec4 removes `inferUnionLiteralSenderType`, `messageLiteralType`, the `getConstSenderType` override, and their `U::Int("bad")` regression test from this branch. Existing stream and union/Switch e2e plus analyzer tests pass without them.

The extracted follow-up is [#1144](https://github.com/nevalang/neva/pull/1144), currently based on #1042 and intended to be retargeted to `main` after #1042 merges.

The only analyzer change still relevant to this PR is resolving the explicit type instantiation in `stream<T>::Tag`. Its raw-generic fallback and the related type-system semantics remain open in the next discussion; this PR no longer claims that the general payload-validation work is required by streams.

internal/compiler/typesystem/subtype_checker.go
emil14: Implemented the narrow fix in 27232945. The earlier subtype change was not needed.

Root cause: the PR had changed `parseConstSenderLiteral` from assigning a union literal its nominal type to assigning the structural literal `{ Tag: nil }`. That artificial representation forced the global rule `{Data:nil} <: {Data:T}`.

The parser now assigns the nominal instantiated type through the new grammar node: `stream<T>::Data` has type `stream<T>`, with its `T` argument preserved. The initial attempt exposed the real analyzer defect: literal sender resolution was the only path in network analysis that did not receive `iface.TypeParams.ToFrame()`, so it could not resolve that `T`. Passing the existing frame through both literal-sender resolution and `Union:tag` validation resolves generic literals normally; the raw `BodyExpr` fallback is removed too.

The subtype checker is strict again: a tag-only structural member cannot match a typed member. No `Switch`/`Union` exception and no new pattern type are introduced here.

Added a parser regression for preserving generic union literal type arguments. Parser/analyzer/typesystem tests, targeted stream and Switch e2e, and `golangci-lint --new-from-rev=origin/main` all pass. General payload validation remains split into #1144.

std/streams/filter.neva
emil14: Rechecked against the old graph implementation. Keeping `SplitCtrl` is justified by the stream protocol, not by a special status of streams.

`Cond<T>` consumes one `data` and one `if` for every iteration. With `T = stream<X>`, `Open` and `Close` would also require predicate booleans, which do not exist. Routing those events around `Cond` creates parallel output paths; without a barrier, `Data` or `Close` can overtake `Open` or an earlier predicate result. The current stdlib has no graph primitive that atomically retains one event, waits for its matching bool, and then serializes it to one of two stream outputs. `Zip` is not equivalent because it buffers streams and changes latency/memory semantics.

`SplitCtrl` is the minimal protocol adapter: it forwards Open/Close to both branches; for Data it sends the payload to the predicate, waits for exactly one bool, routes the original Data event, and only then reads the next event. That explicitly preserves event order and backpressure.

Added `e2e/stream_split` in bb6bcd84. It splits `[1,2,3,4]`, materializes both outputs only after Close, and asserts `else=[1,3]`, `then=[2,4]`; this covers branch lifecycle and order in addition to payload routing.

std/streams/for_each.neva
emil14: I agree that `Controller` should not become a general stdlib design pattern. The narrow conclusion for this PR is to keep these three private runtime protocol adapters, not to promote or generalize them.

They all enforce the same essential stream invariant: at most one Data event is in flight while a scalar user handler produces its continuation. ForEach waits for `done` then forwards the original event; Map waits for `mapped` then emits a new Data event; Split waits for the predicate result then chooses one original-event output. Open/Close are forwarded as protocol events. This state cannot be represented by the current graph primitives without either losing ordering/backpressure or introducing a more general continuation/barrier primitive.

A single Go mega-controller/helper would hide those distinct output and error semantics behind callbacks without materially removing the state machines, so it would be less clear rather than more reusable. The right boundary is therefore private adapters that explicitly document their protocol contract.

Existing e2e cover the behavior: `slow_iteration_with_for` asserts ordered sequential output 1..10; `slow_iteration_with_map` and `map_list_verbose` assert ordered mapped output, including repeated execution. `stream_split` now covers the third variant. I will not add a public controller abstraction in this PR; a future language/runtime design pass can revisit whether a first-class graph-level continuation primitive can replace this family.

benchmarks/atomic/streams_map/main.neva
emil14: So how is this possible this benchmark is still having multiple imports?

UPD: Ok I see where this coming from. Read other comments and u'll fogure out what to do about this

docs/style_guide.md
```diff
@@ -22,6 +22,9 @@ Use tabs over spaces.
 
 ### Imports
 
+Always use multiline import blocks, even for a single import. Do not separate
```
emil14: Since it's not supported by parser, we should not even talk about this

pkg/ast/flowast.go
```diff
@@ -90,8 +90,9 @@ func (pkg Package) Entities() func(func(EntitiesResult) bool) {
 }
 
 type File struct {
-	Imports  map[string]Import `json:"imports,omitempty"`
-	Entities map[string]Entity `json:"entities,omitempty"`
+	Imports      map[string]Import `json:"imports,omitempty"`
+	Entities     map[string]Entity `json:"entities,omitempty"`
+	ImportBlocks []core.Meta       `json:"-"`
```
emil14: I don't think we should make AST more complicated because of grammar changes related to imports. It was completely fine before.

internal/runtime/stream.go
```diff
@@ -0,0 +1,79 @@
+package runtime
```
emil14: Right now honestly I don't think it's good idea to have stream representation at the level of the runtime package. Stream is just a union specific case for runtime, no need to make runtime more complex because of that. Stream must exist at the level of runtime funcs only. Create stream_helpers.go and stream_helpers_test.go in runtime/funcs and that's it. Move all this stuff there and highlight the fact that streams are JUST unions. It's just the thing about them, that they appear a lot, which makes them unions that we see a lot. So for convinience we have helpers for such unioons

internal/runtime/funcs/ports_helpers.go
```diff
@@ -0,0 +1,25 @@
+package funcs
```
emil14, line 1: it should be called ports_helpers.go instead

internal/runtime/funcs/stream_int_range.go
```diff
@@ -8,69 +8,83 @@ import (
...
-	fromIn, err := io.In.Single("from")
+func (rangeInt) Create(runtimeIO runtime.IO, _ runtime.Msg) (func(ctx context.Context), error) {
+	fromIn, err := singleInport(runtimeIO, "from")
 	if err != nil {
-		//nolint:wrapcheck // TODO(strict-lint phase 1): temporary suppression; remove after strict cleanup.
 		return nil, err
 	}
 
-	toIn, err := io.In.Single("to")
+	toIn, err := singleInport(runtimeIO, "to")
```
emil14, line 17: you sure we need these helpers? is there a reason to make runtime more high level maybe?

internal/runtime/funcs/stream_just.go
```diff
@@ -0,0 +1,40 @@
+package funcs
+
+import (
+	"context"
+
+	"github.com/nevalang/neva/internal/runtime"
+)
+
+type streamJust struct{}
```
emil14, line 9: Shouldn't this be streamItem?

internal/compiler/parser/parser_test.go
```diff
@@ -89,6 +89,51 @@ func TestParser_ParseFile_PortlessArrPortAddr(t *testing.T) {
...
+			fmt
+		}
+
+		import {
+			streams
+		}
+	`))
+
+	require.Nil(t, err)
+	require.Len(t, file.ImportBlocks, 2)
```
emil14: there must be only one import block per file

internal/compiler/parser/neva.g4
```diff
@@ -17,9 +17,9 @@ compilerDirective: HASH IDENTIFIER compilerDirectivesArg?;
 compilerDirectivesArg: LPAREN IDENTIFIER RPAREN;
 
 // Imports
-importStmt: IMPORT NEWLINE* LBRACE NEWLINE* importBlockItem* RBRACE;
-importBlockItem: (importDef | COMMENT) NEWLINE*;
-importDef: importAlias? importPath (COMMA)? COMMENT? NEWLINE*;
+importStmt: IMPORT NEWLINE* LBRACE NEWLINE+ importBlockItem* RBRACE;
+importBlockItem: (importDef | COMMENT) NEWLINE+;
+importDef: importAlias? importPath COMMENT?;
```
emil14: IMO imports were fine, it's just that at the parsing level we should've restrict to single import statement per file

internal/compiler/analyzer/analyzer.go
```diff
@@ -157,9 +157,14 @@ func (a Analyzer) analyzePkg(pkg src.Package, scope src.Scope) (src.Package, *co
 	// preallocate
 	analyzedFiles := make(map[string]src.File, len(pkg))
 	for fileName, file := range pkg {
+		if err := a.validateFileImports(fileName, file, *scope.Location()); err != nil {
```
emil14: jfyi import statements count validatin should live in parser, it has not much to do with semantics IMO

internal/compiler/analyzer/const_test.go
```diff
@@ -55,6 +55,27 @@ func TestAnalyzeConstStringLiteralAllowed(t *testing.T) {
 	require.Nil(t, err)
 }
 
+func TestAnalyzePkgRejectsMultipleImportStatements(t *testing.T) {
```
emil14: how is this related to constants?

std/streams/streams.neva
```diff
@@ -27,19 +43,21 @@ pub def FromDict<T>(data dict<T>) (res stream<DictEntry<T>>)
...
 pub def Range(from int, to int) (res stream<int>)
 
+// Enumerate attaches a 0-based index to each Data item in a stream.
+#extern(stream_enumerate)
+pub def Enumerate<T>(data stream<T>) (res stream<Enumerated<T>>)
+
 // --- Synchronization ---
 
-// Wait blocks until last stream item arrive, then sends a signal.
+// Wait blocks until stream close item arrives, then sends a signal.
```
emil14: Until stream closes* is a better description IMO

std/streams/for_each.neva
```diff
@@ -1,47 +1,34 @@
...
+// It uses backpressure to ensure the handler processes items sequentially:
+// the next Data item is not forwarded until the handler finishes the previous one.
+// This prevents concurrency issues in side-effecting handlers.
 //
-// Like other iterators, it processes items one at a time without blocking the stream.
-// To wait for all items to be processed, use with `streams.Wait`.
+// To wait for all items to be processed, use it with streams.Wait.
 pub def ForEach<T>(data stream<T>) (res stream<T>, err error) {
-    // IDEA - if we use sync.Step, we don't have to mess with indexes
+	controller ForEachController<T>
```
emil14: lets call controller nodes just ctrl and components Controller* (here ForEachController is ok (if we keep controllers at all))

std/streams/filter.neva
```diff
@@ -1,29 +1,35 @@
...
+// SplitCtrl preserves stream ordering while routing Data items by predicate.
+//
+// @inport data Input stream.
+// @inport pass Predicate result for the item emitted on item.
+//
+// @outport item Payload to test with the predicate.
+// @outport then Stream containing items accepted by the predicate.
+// @outport else Stream containing items rejected by the predicate.
+#extern(stream_split_controller)
+def SplitCtrl<T>(data stream<T>, pass bool) (item T, then stream<T>, else stream<T>)
```
emil14: SplitCtrl -> SplitController

std/streams/filter.neva
```diff
@@ -1,29 +1,35 @@
...
-    predicate IPredicate<T>
-    ---
-    :data -> [cond:data, .data -> predicate]
-    predicate -> cond:if
-    cond:then -> :then
-    cond:else -> :else
+	ctrl SplitCtrl<T>
+	predicate IPredicate<T>
+	---
+	:data -> ctrl:data
```
emil14: remember omit port names where possible

benchmarks/atomic/streams_map/main.neva
emil14: Fixed in b1fecb6c. The benchmark no longer depends on an analyzer-only check: the parser grammar now permits exactly one import statement per file. Existing import syntax remains valid; a second `import { ... }` is rejected during parsing.

docs/style_guide.md
emil14: Removed in b1fecb6c. The style guide no longer prescribes multiline-only or comma-free import syntax, because that was an unnecessary language restriction rather than an established parser contract.

pkg/ast/flowast.go
emil14: Agreed and fixed in b1fecb6c. `ImportBlocks` has been removed from `ast.File`; import-count tracking no longer adds AST metadata. The grammar itself accepts at most one `importStmt`, so duplicate imports fail before the AST is built.

internal/runtime/stream.go
emil14: Moved in b1fecb6c. `internal/runtime/stream.go` and its tests are gone. The helpers now live privately in `internal/runtime/funcs/stream_helpers.go` and construct/check ordinary `runtime.UnionMsg` values with `Open`, `Data`, and `Close` tags. This keeps streams as a commonly used union convention, not a runtime-level data model.

internal/runtime/funcs/ports_helpers.go
emil14: Renamed in b1fecb6c: `ports.go` is now `ports_helpers.go`.

internal/runtime/funcs/stream_int_range.go
emil14: I kept the range helpers. They are private decomposition of one generator protocol: emit Open, emit the ascending or descending Data sequence, then emit Close. They do not make `runtime` higher-level or create reusable runtime API; the stream-specific union helpers are now private to `runtime/funcs`. The split keeps the three control-flow cases readable and makes the Close-on-every-successful-range invariant explicit.

internal/runtime/funcs/stream_just.go
emil14: Yes. Renamed in b1fecb6c: `stream_just.go` is now `stream_item.go`, its implementation is `streamItem`, and the stdlib binding is now `#extern(stream_item)`, matching the public `streams.Item` API.

internal/compiler/parser/parser_test.go
emil14: Implemented in b1fecb6c. `TestParser_ParseFile_RejectsMultipleImportStatements` now asserts the parser rejects two import blocks. Companion parser cases assert that the pre-existing single-line and comma-separated forms remain valid.

internal/compiler/parser/neva.g4
emil14: Implemented in b1fecb6c. I restored the previous import-block grammar and moved only the one-statement constraint into `prog`: it permits zero or one `importStmt` plus any non-import statements. A second import statement is therefore a parse error, with no AST or analyzer policy required.

internal/compiler/analyzer/analyzer.go
emil14: Agreed and fixed in b1fecb6c. Import-count validation was removed from `Analyzer`; the parser grammar owns it now. `analyzer.go`, `const_test.go`, and `ast.File` no longer contain import-block bookkeeping.

internal/compiler/analyzer/const_test.go
emil14: Agreed and fixed in b1fecb6c. The unrelated analyzer test was removed along with analyzer-level import validation. The regression is now in `internal/compiler/parser/parser_test.go`, where the syntax rule belongs.

std/streams/streams.neva
emil14: Updated in b1fecb6c: the comment now says `Wait blocks until stream closes, then sends a signal.`

std/streams/for_each.neva
emil14: Applied in b1fecb6c. Internal controller node aliases are now `ctrl` in `ForEach` and `Map`; the component names remain descriptive (`ForEachController`, `MapController`).

std/streams/filter.neva
emil14: Renamed in b1fecb6c: `SplitCtrl` is now `SplitController`, while its internal node alias remains `ctrl`.

std/streams/filter.neva
emil14: Applied where the endpoint is unambiguous. In this connection, `ctrl` has multiple inports (`data`, `pass`), so `:data -> ctrl` is rejected by the analyzer; `ctrl:data` is required. The single-inport `Filter` connection already uses the short form `:data -> split`.

internal/compiler/parser/neva.g4
```diff
@@ -2,11 +2,13 @@ grammar neva;
 
 /* PARSER */
 
-prog: (NEWLINE | COMMENT | stmt)* EOF;
+prog:
```
emil14: keep it simple at the grammar level, just validate in the listener helpers level and throw readable error

std/streams/streams.neva
```diff
@@ -15,6 +20,17 @@ pub def FromList<T>(data list<T>) (res stream<T>)
...
+//
+// This is useful when an API expects a stream but the producer has one value,
+// for example when passing a single pixel to image.New or adapting scalar data
+// to stream combinators.
+//
+// @inport data Value to emit as the only Data item.
+//
+// @outport res Stream that emits Open, Data(data), and Close.
+#extern(stream_item)
+pub def Item<T>(data T) (res stream<T>)
```
emil14: - Image.New is bad example IMO
- Is "Item" genuinely good name? Maybe "Once" good do a better job? Like "stream.Once"? Or maybe the old one "Just" was a better name actually? Please think about this one deeply.

std/streams/map.neva
```diff
@@ -1,19 +1,26 @@
...
+	ctrl MapController<T, Y>
+	handler IMapHandler<T, Y>
+	---
+	:data -> ctrl:data
+	ctrl:item -> handler
+	handler -> ctrl:mapped
+	ctrl:res -> :res
 }
+
+// MapController serializes item mapping and stream event forwarding.
```
emil14: "serializes item mapping and stream event forwarding" not good enough comment IMO, its not immidiately clear for me what that means exactly

internal/runtime/funcs/stream_item.go
```diff
@@ -0,0 +1,40 @@
+package funcs
+
+import (
+	"context"
+
+	"github.com/nevalang/neva/internal/runtime"
+)
+
+type streamItem struct{}
```
emil14: if we will reename at neva stdlib level then we shoul rename here too as well

internal/compiler/typesystem/subtype_checker.go
emil14: So why we still have changes to typesystem? This file specifically?

internal/runtime/funcs/stream_enumerate.go
```diff
@@ -0,0 +1,63 @@
...
+	switch {
+	case isStreamOpen(msg):
+		*idx = 0
+		return resOut.Send(ctx, newStreamOpenMsg())
+	case isStreamData(msg):
+		item := runtime.NewStructMsg([]runtime.StructField{
+			runtime.NewStructField("idx", runtime.NewIntMsg(*idx)),
+			runtime.NewStructField("item", streamDataValue(msg)),
+		})
+		if !resOut.Send(ctx, newStreamDataMsg(item)) {
```
emil14, line 53: why are we creating runtime struct msg just to create a union? feels weird...

internal/runtime/funcs/registry.go
```diff
@@ -54,7 +54,12 @@ func NewRegistry() map[string]runtime.FuncCreator {
 		"dict_to_stream":       dictToStream{},
 		"stream_to_dict":       streamToDict{},
 
-		"stream_int_range": rangeInt{},
+		"stream_int_range":           rangeInt{},
```
emil14: adjust names for consistency, e.g. streamIntRange

internal/runtime/funcs/stream_for_each_controller.go
```diff
@@ -0,0 +1,70 @@
...
+	"context"
+
+	"github.com/nevalang/neva/internal/runtime"
+)
+
+// streamForEachController serializes stream item handling for ForEach.
+// It forwards Open/Close immediately and forwards Data only after done signal.
+type streamForEachController struct{}
+
+func (streamForEachController) Create(runtimeIO runtime.IO, _ runtime.Msg) (func(ctx context.Context), error) {
```
emil14: about these controllers (map controller, for each controller, etc.)

can't we implement them using existing neva components such as switch, etc.?

why create them in go?

internal/compiler/parser/neva.g4
emil14: Agreed and changed in 5e48d148. The grammar is back to the simple `prog: (NEWLINE | COMMENT | stmt)* EOF` form. `treeShapeListener.EnterImportStmt` calls the listener helper `validateImportStmt`; it rejects the second block with the readable parser error `file must contain at most one import statement`, located at that block. No AST or analyzer state is involved.

std/streams/streams.neva
emil14: I reconsidered this and kept `Item` in 5e48d148, while removing the Image.New example. `Item` states the concrete value-level contract: this source turns one item into a stream. `Once` suggests a temporal/subscription policy and does not say what is emitted; `Just` is recognizable in FP/reactive ecosystems but imports terminology that the rest of the stdlib does not use. `Item` also composes naturally with `FromList`, `FromDict`, and `FromString` as a source family. Since the public name remains `Item`, the runtime name stays `streamItem`/`stream_item` consistently.

std/streams/map.neva
emil14: Expanded in 5e48d148. The comment now states the actual protocol: Open/Close are forwarded directly; for each Data event the controller sends its payload to the mapper, waits for exactly one mapped value, emits the corresponding Data event, and only then accepts the next input event. This is the ordering/backpressure contract the controller implements.

internal/runtime/funcs/stream_item.go
emil14: The names now agree. The public stdlib entity remains `streams.Item`; its runtime binding is `stream_item`, its implementation is `streamItem`, and the source file is `stream_item.go`. I did not rename further because, after reconsidering the API naming discussion, `Item` is the selected public term.

internal/runtime/funcs/stream_enumerate.go
emil14: `Enumerated<T>` is a struct `{ idx int, item T }`, while every stream event is a union message `stream::Data(payload)`. Therefore the runtime must first encode the Enumerated value as `runtime.StructMsg`, then use that value as the Data union payload. It is not creating two competing representations: the struct is the item value, and the union is the stream-protocol envelope. I added that intent comment next to the construction in 5e48d148.

internal/runtime/funcs/registry.go
emil14: Adjusted in 5e48d148: `rangeInt` is now `streamIntRange`, and `range_int.go` is now `stream_int_range.go`. The stream runtime creators now use the consistent `stream*` implementation-name family.

internal/runtime/funcs/stream_for_each_controller.go
emil14: Not with the currently available Neva components while preserving the required protocol. `Switch`/`Cond` can select based on messages, but ForEach/Map/Split must retain one original Data event, invoke a scalar handler, wait for its one continuation (`done`, mapped value, or predicate bool), then forward or rebuild exactly that event before accepting the next Data. Open/Close must bypass the scalar handler without overtaking any Data result. A pure graph built from existing nodes either sends the event and continuation down parallel paths with no barrier/order guarantee, or applies the handler to Open/Close too. `Cond` specifically needs a data and condition on every iteration, whereas Open/Close intentionally have no predicate/mapper result. These private Go controllers are therefore protocol adapters, not language-level special cases. Replacing them cleanly would require a general graph-level continuation/barrier primitive; that is a separate runtime/language design task, not an implementation obtainable from the existing primitives.

internal/runtime/funcs/stream_item.go
emil14: I decided it must be stream_just/streamJust

std/streams/streams.neva
emil14: think FP terminology can be ok especially given its actually used already. Im pretty sure we have "Tap"? Just is better IF its really what is used for in FP world. Neva is indeed not much far from functional programming

emil14: Agreed. `Just` is established terminology in reactive APIs for a source that emits exactly one value and then completes; it describes this constructor more precisely than `Item`. FP vocabulary is already appropriate in Neva (`Tap` is a good example), so I renamed the public API in a78e64aa: `streams.Item<T>` is now `streams.Just<T>`.

internal/runtime/funcs/stream_item.go
emil14: Done in a78e64aa. The binding is now `#extern(stream_just)`, the registry key is `stream_just`, the implementation is `streamJust`, and the file is `stream_just.go`. The examples and the imported-type-argument e2e now use `streams.Just` as well.

internal/compiler/typesystem/subtype_checker.go
emil14: You are right: there should be no change in this file. It contained only an earlier broad lint-driven refactor, not any required stream or generic-union semantics. I removed the entire `subtype_checker.go` diff in a78e64aa; its final content is byte-for-byte the `main` version. The remaining compiler fix is limited to: parsing `typeInstExpr::Tag`, preserving the nominal instantiated union type, and passing the existing generic type frame into analyzer resolution. No subtype rule, Switch/Union exception, or pattern-type mechanism is introduced by this PR.

internal/runtime/funcs/accumulator.go
```diff
@@ -2,50 +2,44 @@ package funcs
...
 
 	"github.com/nevalang/neva/internal/runtime"
 )
 
 type accumulator struct{}
 
-//nolint:cyclop,gocognit,gocyclo,varnamelen // TODO(strict-lint phase 1): temporary suppression; remove after strict cleanup.
-func (a accumulator) Create(io runtime.IO, _ runtime.Msg) (func(ctx context.Context), error) {
-	initIn, err := io.In.Single("init")
+//nolint:cyclop,funlen,gocognit,gocyclo // The stream cycle and its cancellation paths must remain together.
```
emil14, line 13: I see a lot of changes in runtime/funcs go files, I dont have time to understand them all so Ill ask. It's critical that (respecting specific semahntics of the runtime function of course) they are deadlock-free.

That is, e.g. when a native component (runtime func) has multiple input ports, it must receive from all concurrently. That's why we have receive2/3/4 helper functions.

This must be a harness rule for the runtime functions. Sometimes it *might* (and I'm not even sure!) make sense to receive sequentially, if specific component logic requires it (which maybe does, including this PR, I dunno), but generally we must keep deadlocks in our mind and do our absolute best to avoid them.

Now carefully analyze changes made in this PR regarding runtime funcs and tell me - are we safe? If not, make changes.

emil14: Audited every runtime function changed by this PR against the multi-input receive invariant and pushed the fixes in 19531546.

Fixed:
- `stream_zip`: Open events and every left/right item pair are now received concurrently. The former left-then-right receives could block the right producer while the left side was unavailable. Once one stream closes, only the still-open stream is drained, so that final drain is intentionally single-input.
- `arr_port_to_stream`: array slots are now received concurrently via `ReceiveAll`, buffered by index, then emitted in deterministic index order.
- The new concurrent `ReceiveAll` path exposed a data race in its shared `success` flag. I fixed it in `runtime/program.go`: each worker now reports its result through the existing buffered result channel, and the caller alone updates `success`.

Already concurrent:
- `stream_int_range`, `strings_join_list`, and the initial `strings_join_stream` data/separator handshake use `receive2`.
- `stream_zip_many` and `stream_product` run one reader per input.
- `accumulator` reads `upd` and `flush` concurrently after its required `init` starts a cycle.

Intentional sequential receives:
- `sync.Turn` must wait for `done` before it receives the next `data`; that backpressure is its public contract and is covered by a unit test.
- After `strings_join_stream` has received the separator for a stream, only `data` remains part of that cycle.
- The remaining changed functions have one input port.

Added focused unit tests proving that `Zip` accepts the right input before the left arrives and that `ArrayPortToStream` accepts a later array slot before an earlier one. Passed:
`go test ./internal/runtime ./internal/runtime/funcs -count=1`
`go test -race ./internal/runtime/funcs -run 'Test(StreamZipReceivesBothInputsConcurrently|ArrayPortToStreamReceivesSlotsConcurrently|TurnWaitsForDoneBeforeReceivingNextData)' -count=1`
`go test ./examples/stream_zip ./examples/stream_zip_many ./examples/stream_product -count=1`
`golangci-lint run ./internal/runtime ./internal/runtime/funcs`.

Issue #1008, opened 2026-01-27
Optimize CI/CD pipeline for test execution (E2E-first)


emil14:
Our current CI test action regularly takes ~14–15 minutes to complete. While this is not catastrophic yet, it is already at the threshold where it noticeably slows down iteration and will only get worse as the project grows.

This issue is about **optimizing the CI/CD pipeline for tests**, with **end-to-end (E2E) tests as the first priority**. This is explicitly *not* about optimizing the language runtime or compiler internals themselves — only about test and CI infrastructure.

## High-level directions to explore

This issue intentionally avoids deep implementation details. The goal is to identify and implement **low-risk, high-ROI improvements** in CI behavior.

**1) Re-enable and validate parallelism for E2E tests**

* Revisit why parallel execution was disabled in the past
* Test whether current E2E isolation (temp dirs, per-test state) is sufficient to avoid interference
* If safe, re-enable parallel execution (threads/processes/workers)

**2) Reuse a prebuilt compiler executable across E2E tests**

* Currently the compiler/toolchain is rebuilt for every E2E test
* Investigate building the compiler once per CI run and reusing it
* Review how `go run` caching behaves in this context and whether we can do better
* Coordinate with the existing issue about compiler reuse and reference it from here

emil14:
Related to https://github.com/nevalang/neva/issues/979

Issue #979, opened 2025-12-27
Improve e2e CLI helper performance by reusing built `neva` binary


emil14:
**Context**
We switched `e2e.Run` to `go build`+exec so the CLI is built from the repo root while tests run from arbitrary working dirs. This fixed module resolution issues but still builds (and links) the binary on every test invocation.

**Problem**
Repeated `go build -o <temp>` in e2e tests adds overhead. Although package compilation is cached, the final link and binary write still occur each run.

**Proposal**
- Cache the built CLI binary across test cases (same process) using `sync.Once` or a package-level lazy build.
- Write the binary to a stable path (e.g., under `os.TempDir()` keyed by `go env GOMOD` + a hash of Go source timestamps) to reuse across test processes when safe.
- Add a fast path to skip rebuild when inputs unchanged; rebuild if source modtime/hash differs.
- Ensure cleanup on hash change or `go clean -cache`.
- Keep current behavior as fallback if caching fails.

**Acceptance criteria**
- e2e tests no longer rebuild/link the CLI binary per test when sources unchanged.
- Module and build caches remain utilized; correctness unchanged.
- Document the caching behavior in `pkg/e2e` helper comments.

Issue #767, opened 2024-11-17
[CI/CD | E2E tests] Parallel execution fails (race-conditions)


emil14:
# Problem

It's race-condition. Here's the [change I've made to CD/CD that makes tests sequential](https://github.com/nevalang/neva/pull/850/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R25)

Now that we know that this is race-condition, we need to fix it.

## Guess

I think each tmp go-mod dir should have unique name. Perhaps using time.Now() to add suffix is the simplest solution. You can check internal/compiler/backend folder.

allain:
Changing the tmp directory generation to use os.MkdirTemp dit not resolve the issue but should have if it was just a matter of the directories clobbering each other.

```golang
// neva/internal/compiler/backend/golang/native/backend.go
tmpGoModuleDir, _ := os.MkdirTemp(output, "tmp")
```

```golang
// neva/internal/compiler/backend/golang/wasm/backend.go
tmpGoProj, _ := os.MkdirTemp(dst, "tmp")
```

I could not find any other references "tmp" in the code base though I may have missed some.

allain:
This is resolved if you change rewriteStdlibOntoDisk to ensueStdlibOnDisk and only make it write the stdlib if it's not already there.

What's happening is the stdlib is getting deleted under tests as they are being run.

PR #1062, opened 2026-03-16
test(ci): split unit/e2e pipeline and cache e2e neva binary

emil14:
## Summary
- add shared `neva` binary cache in `pkg/e2e` using a stable cache path and file fingerprint (`path + size + mtime`)
- keep safe fallback to per-test build if cache path/fingerprint/locking fails
- guard `ensureStdlib()` in `builder.New` with existing file lock to avoid concurrent stdlib rewrite races
- split CI test workflow into ordered jobs: `lint -> unit_tests -> e2e_tests`
- run `e2e/examples` with conservative package parallelism (`-p 2`)
- disable duplicate lint run on PRs in `lint.yml` (lint is now in `test.yml` pipeline)

## Why
- addresses test runtime and CI pipeline concerns from:
  - #1008 (CI test optimization)
  - #979 (avoid rebuilding neva binary for each e2e package)
  - #767 (stepwise return of package parallelism with race-risk mitigation)

## Validation
- `go test ./pkg/e2e ./internal/builder/...`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.5.0 run ./pkg/e2e ./internal/builder/...`
- `go test -count=1 -v ./e2e/hello_world_verbose ./examples/hello_world`
- `go list ./... | grep -Ev '^github.com/nevalang/neva/(e2e|examples)(/|$)' | xargs -r go test -run '^$' -count=1`
- `go test -run '^$' -count=1 -p 2 ./e2e/... ./examples/...`


pkg/e2e/e2e.go
```diff
@@ -225,3 +245,267 @@ func buildNevaBinary(t *testing.T, repoRoot, mainPath string) string {
...
+	if err := json.Unmarshal(raw, &names); err != nil {
+		return fmt.Errorf("decode go list file set for %s: %w", dir, err)
+	}
+
+	addPackageFiles(filesSet, dir, names)
+
+	return nil
+}
+
+// acquireCacheLock provides a simple file lock used by concurrent test binaries.
```
emil14: comments needs to be more descriptive, they must explain semantics of the function, e.g. here how timeout works, what is func() returned, etc. upd AGENTS.md as well as code

pkg/e2e/e2e.go
```diff
@@ -205,6 +211,20 @@ func getExitCode(err error) int {
 func buildNevaBinary(t *testing.T, repoRoot, mainPath string) string {
 	t.Helper()
 
+	binPath, err := buildNevaBinaryFromCache(repoRoot, mainPath)
```
emil14: add to agents.md - one-liner comments are encouraged for code blocks with more than 3 lines, when code is generated from scratch (and upd this PR)

pkg/e2e/e2e.go
```diff
@@ -225,3 +245,267 @@ func buildNevaBinary(t *testing.T, repoRoot, mainPath string) string {
...
 }
+
+// buildNevaBinaryFromCache resolves a reusable neva binary from a process-safe cache.
+func buildNevaBinaryFromCache(repoRoot, mainPath string) (string, error) {
+	cacheRoot, err := e2eCacheRootDir()
+	if err != nil {
+		return "", fmt.Errorf("resolve e2e cache root: %w", err)
+	}
+
+	cacheKey, err := nevaBuildFingerprint(repoRoot)
```
emil14: is there anything we could improve about how stdlib fingerprint works? e.g. maybe we are reading the whole stdlib to get hash while we can do the same stuff we do here? maybe we can even move fingerprint logic to pkg layer and reuse?

PR #1100, opened 2026-04-09
benchmarks: add routing complex baseline and observability surface

emil14:


Makefile
```diff
@@ -49,6 +49,10 @@ nilaway:
 		-include-pkgs="github.com/nevalang/neva/internal" \
 		./...
 
+.PHONY: bench-runtime
+bench-runtime:
+	go test ./benchmarks -run=^$$ -bench BenchmarkRuntimeE2E -benchtime=1x -count=1
```
emil14: explain `-run=^$$ -bench BenchmarkRuntimeE2E -benchtime=1x -count=1`

1. shouldn't we run everything under benchmarks folder?
2. what is -benchtime=1x?
3. are you sure about -count=1? don't we wanna parallelism? tests are parallel already AFAIK

docs/README.md
```diff
@@ -7,4 +7,5 @@ Welcome to Nevalang documentation!
 - **[Tutorial](./tutorial.md)** - Best place to start.
 - **[Style Guide](./style_guide.md)** - Best practices to follow.
 - **[Questions and Answers](./qna.md)** - If you curious why things exactly the way they are.
-- **[The Book](./book/README.md)** - A deep dive into the language and how it works. Make sure you've read tutorial!
\ No newline at end of file
+- **[The Book](./book/README.md)** - A deep dive into the language and how it works. Make sure you've read tutorial!
+- **[Runtime Benchmarks](../benchmarks/README.md)** - Baseline suite for runtime comparisons and benchmark workflow usage.
```
emil14: these docs are for langauge users, not language devs

benchmarks/README.md
```diff
@@ -0,0 +1,24 @@
+# Runtime Benchmarks
```
emil14: just make sure we not duplicating context with AGENTS.md, AGENTS file is enough tho

Makefile
emil14: Addressed: kept a single `bench-runtime` target, switched scope to `./benchmarks/...`, and documented why `-run=^$`, `-benchtime=1x`, and `-count=1` are used for this external-process benchmark harness.

docs/README.md
emil14: Addressed: removed the runtime benchmark link from `docs/README.md` so language-user docs stay focused.

benchmarks/README.md
emil14: Addressed: trimmed `benchmarks/README.md` to a minimal pointer to `benchmarks/AGENTS.md` plus the single local run command, to avoid duplicating durable guidance.

emil14: remove this file and let's document that sub-directories usually do not need README.md if there's AGENTS.md already presented

.github/workflows/runtime-benchmarks.yml
```diff
@@ -0,0 +1,63 @@
...
+  release:
+    types:
+      - published
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  runtime_benchmarks:
```
emil14: do we have any other benchmarks that this? shouldnt we just call it benchmarks?

Makefile
```diff
@@ -49,6 +49,15 @@ nilaway:
 		-include-pkgs="github.com/nevalang/neva/internal" \
 		./...
 
+.PHONY: bench-runtime
+# Bench suite runner:
+# - `./benchmarks/...` keeps scope limited to benchmark harness packages.
+# - `-run=^$$` skips unit tests and executes only benchmark functions.
+# - `-benchtime=1x` runs each benchmark exactly once (external process workload).
```
emil14: > runs each benchmark exactly once (external process workload).

i dont uynderstand this, explain, should we run benchmarks multimple times? b.Count or whatever that called?

Makefile
```diff
@@ -49,6 +49,15 @@ nilaway:
...
+# - `-count=1` disables repeat runs to keep CI/runtime snapshots deterministic.
```
emil14: what is repeat runs and why dont we need it? how is this working together with -benchtime? does it have to do anything with parallelism?

Makefile
```diff
@@ -49,6 +49,15 @@ nilaway:
...
+bench-runtime:
```
emil14: maybe just benchmarks? like all of them, in the whole repo, including ones we going to have for runtime/funcs?

benchmarks/complex/selectors_routers/main.neva
```diff
@@ -0,0 +1,66 @@
...
+	wait streams.Wait
+	---
+	:start -> [
+		1 -> range:from,
+		10000 -> range:to
+	]
+	range -> map_complex -> wait -> :stop
+}
+
+def ComplexFlow(data int) (res int) {
```
emil14, line 27: In this PR, also in all other benchmarks here in benchmarks directory, in *.neva files, add at least one line of doc-comment to each component, and update docs/style-guide.md if it's not there with instruction that if component is more complex than 3 nodes and/or 3 connections it's encouraged to have at least single line of doc-comment explaining what it does. This must not reduce necessarily of having good naming though. Also update cursor/rules for *.neva files - add instruction that for every AI-generated entity, newly added one in the sesison, it's required to have at least single-line doc comment that explains what this entity does.

Makefile
emil14: Clarified and implemented as requested: `-benchtime=1x` and `-count=1` are both set for one-pass snapshot runs; they are not a parallelism control. Parallelism remains Go benchmark default behavior.

Makefile
emil14: Applied. There is now a single `make benchmarks` target. It runs benchmark functions across repo packages (excluding `e2e/examples` modules), so future benchmark additions are covered by default.

.github/workflows/runtime-benchmarks.yml
emil14: Updated: workflow and job are now named `benchmarks`.

Makefile
emil14: Updated Make target to `benchmarks` and retained benchmark-only execution flags. Current command runs benchmark functions across repository packages (excluding `e2e/examples` modules) with `-run=^$ -bench . -benchtime=1x -count=1`.

benchmarks/complex/selectors_routers/main.neva
emil14: Applied in this push: added single-line doc comments across benchmark `.neva` entities, added the style-guide note for components over 3 nodes/connections, and added the Neva authoring rule requiring doc comments for newly added AI-generated entities.

benchmarks/README.md
emil14: Applied: `benchmarks/README.md` remains removed, and `benchmarks/AGENTS.md` now explicitly states to avoid subdirectory `README.md` when an `AGENTS.md` already exists.

Makefile
emil14: Final state in latest push: single `make benchmarks` target, using benchmark-only flags and a single-pass run (`-run=^$ -bench . -benchtime=1x -count=1`) over repo packages while excluding `e2e/examples` modules.

benchmarks/atomic/builtin_fan_in_streams/main.neva
```diff
@@ -1,5 +1,6 @@
 import { sync }
 
+// Main runs the benchmark scenario.
```
emil14: Oh now that's not what I meant... ask yourself is this a useful comment? common you can do better than that! Example: "Main sends start msg into 2 inports of the fan_in node and awaits until both of them are sent to the output". You can do better than me though, I just tried and was lazy. Here and in all other neva files in this PR. Also update either neva rules in cursor or AGENTS.md - comments must always be useful. Expand on this in there

benchmarks/complex/selectors_routers/main.neva
```diff
@@ -0,0 +1,69 @@
+// Benchmarks a more complex mixed control-flow workload.
+// Combines selectors, routers, and union wrapping/unwrapping.
+import {
+	runtime
+	streams
+}
+
+// Number defines benchmark data used by this scenario.
```
emil14: Another example of useless comment. For example this is not just a number, it's a union. There's a reason why it's a union right? And why does it have only one field, and so one and so forth.

benchmarks/atomic/builtin_fan_in_streams/main.neva
emil14: Applied across the full PR in commit `5e1e133e`: replaced generic comments with behavior-specific descriptions in every touched benchmark `.neva` file (what message flow/path is exercised and what is being measured). Also tightened comment quality rules in `benchmarks/AGENTS.md`, `.claude/rules/neva-files.md`, and `docs/style_guide.md` to explicitly disallow placeholder comments.

benchmarks/complex/selectors_routers/main.neva
emil14: Agreed and fixed. Union/struct comments now explain why those types exist in the benchmark (forcing explicit tagged-union box/unbox and Struct+Union interaction), not just restating names. I applied this quality bar consistently across all benchmark files in this PR.

PR #1111, opened 2026-04-20
runtime/funcs: remove nolint suppressions from operator creators

emil14:
## Summary
This cleanup removes `//nolint` suppressions from the runtime operator creators in `internal/runtime/funcs` by eliminating duplicated creator boilerplate and centralizing port resolution/error handling.

## `nolint` comments removed
Removed **176** `//nolint` comments across **32 operator files** in `internal/runtime/funcs`:
- `dupl`: 24 removals
- `wrapcheck`: 92 removals
- `varnamelen`: 60 removals

Touched operator files:
- `and.go`, `or.go`, `eq.go`, `ne.go`
- `int_add.go`, `int_sub.go`, `int_mul.go`, `int_div.go`, `int_mod.go`, `int_pow.go`
- `int_bitwise_and.go`, `int_bitwise_or.go`, `int_bitwise_xor.go`, `int_bitwise_lsh.go`, `int_bitwise_rsh.go`
- `int_is_greater.go`, `int_is_lesser.go`, `int_is_greater_or_equal.go`, `int_is_lesser_or_equal.go`
- `float_add.go`, `float_sub.go`, `float_mul.go`, `float_div.go`
- `float_is_greater.go`, `float_is_lesser.go`
- `string_add.go`, `string_is_greater.go`, `string_is_lesser.go`
- `int_inc.go`, `int_decr.go`, `int_neg.go`, `float_neg.go`

## Root causes fixed
- Introduced shared operator creator helpers in `internal/runtime/funcs/operator_helpers.go`:
  - `createBinaryFuncConcurrent`
  - `createBinaryFuncSequential`
  - `createUnaryFunc`
  - wrapped port lookup helpers (`getSingleInport`, `getSingleOutport`)
- Removed repeated per-file wiring/loop code that was triggering `dupl`.
- Replaced direct passthrough external errors with wrapped contextual errors at a single shared boundary (`resolve inport/outport ...`), addressing `wrapcheck` at the source.
- Replaced many short-scope/short-name local variables by centralizing logic in helpers, addressing `varnamelen` without adding new suppressions.

## Tests added/updated
Added `internal/runtime/funcs/operators_baseline_test.go` with behavior-locking tests before/after refactor:
- `TestBinaryOperatorsBehavior`
- `TestUnaryOperatorsBehavior`

These cover representative operators from all refactored groups (arithmetic, comparisons, boolean logic, bitwise, unary transforms).

## Checks run
- `go test ./internal/runtime/funcs -run 'TestBinaryOperatorsBehavior|TestUnaryOperatorsBehavior'` (baseline pre-refactor and post-refactor)
- `go test ./internal/runtime/funcs`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest run ./internal/runtime/funcs/...`

## Behavior
No intentional behavior changes; this is a structural cleanup + error-context improvement while preserving runtime operator semantics.


internal/runtime/funcs/operator_helpers.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	return func(ctx context.Context) {
+		for {
+			var leftValue runtime.Msg
+			var rightValue runtime.Msg
+			var leftAvailable bool
+			var rightAvailable bool
+
+			var waitGroup sync.WaitGroup
```
emil14: call it just wg

internal/runtime/funcs/operator_helpers.go
```diff
@@ -0,0 +1,133 @@
...
+	"sync"
+
+	"github.com/nevalang/neva/internal/runtime"
+)
+
+type binaryOperator func(left runtime.Msg, right runtime.Msg) runtime.Msg
+
+type unaryOperator func(input runtime.Msg) runtime.Msg
+
+func createBinaryFuncSequential(runtimeIO runtime.IO, apply binaryOperator) (func(context.Context), error) {
```
emil14: Only createBinaryFuncConcurrent should be present, input arguments must ALWAYS (you can add this to *.go rules under .cursor or whatever they are, as rule for runtime functions) be received concurrently whenever possible, sequential reading might lead to deadlock in neva programs

You can even create p1 issue in gh about this - check all runtiume funcs for this

internal/runtime/funcs/operators_baseline_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+	ctx, cancel := context.WithCancel(context.Background())
+	done := make(chan struct{})
+	go func() {
+		handler(ctx)
+		close(done)
+	}()
+
+	leftInput <- runtime.OrderedMsg{Msg: left}
+	rightInput <- runtime.OrderedMsg{Msg: right}
```
emil14: we could use different order 50/50, to check cuncurrent reading

internal/runtime/funcs/operators_baseline_test.go
```diff
@@ -0,0 +1,189 @@
...
+		"right": runtime.NewInport(nil, runtime.NewSingleInport(rightIn, runtime.PortAddr{Path: "test/in", Port: "right"}, interceptor)),
+	})
+	outports := runtime.NewOutports(map[string]runtime.Outport{
+		"res": runtime.NewOutport(runtime.NewSingleOutport(runtime.PortAddr{Path: "test/out", Port: "res"}, interceptor, resultOut), nil),
+	})
+
+	return runtime.IO{In: inports, Out: outports}, leftIn, rightIn, resultOut
+}
+
+func newUnaryRuntimeIO() (runtime.IO, chan runtime.OrderedMsg, chan runtime.OrderedMsg) {
```
emil14, line 200: we should have tests for each runtime function, this file is not replacement for that, we can have baseline here but should add unit tests for each runtime func including each operator. it does not have to be done in this PR tho, jsut add comments

internal/runtime/funcs/operator_helpers.go
emil14: Good catch on naming. I tried `wg`, but current `varnamelen` lint in this repo rejects it in this scope; kept a lint-compliant name (`group`) to avoid adding suppressions.

internal/runtime/funcs/operator_helpers.go
emil14: Updated: all binary operator creators now use `createBinaryFuncConcurrent` only; sequential helper path was removed from `operator_helpers.go`. Also added a rule in `.claude/rules/go-files.md` to keep binary runtime funcs concurrent by default. Opened follow-up issue for wider runtime-func audit: #1114.

emil14: Correction: follow-up audit issue is #1113.

internal/runtime/funcs/operators_baseline_test.go
emil14: Updated baseline test helper to execute both send orders (`left->right` and `right->left`) and switched binary test inputs to unbuffered channels with timeout guards, so sequential receive regressions are detected.

internal/runtime/funcs/operators_baseline_test.go
emil14: Agreed. Added an explicit TODO note in `operators_baseline_test.go` clarifying this baseline file is not a replacement for per-runtime-function unit tests, and created follow-up tracking issue #1113.

PR #1127, opened 2026-05-16
runtime: typed scalar list/dict containers + hotpath fast-paths

emil14:
## Summary

This PR makes scalar `list<T>` and `dict<T>` values use typed runtime storage while preserving the public Neva types and the untyped fallback for heterogeneous or non-scalar values.

- add `ListMsg` and `DictMsg` container contracts and scalar constructors;
- emit typed constructors for homogeneous scalar list/dict literals in the Go backend;
- preserve typed storage in `list_at`, `list_slice`, `string_join`, and `get(dict)`;
- make a missing key in every scalar typed dict return the normal error instead of falling through to the untyped representation;
- add focused unit tests, construction/traversal/runtime-function benchmarks, and developer guidance for typed-first handling and boxing boundaries.

The unrelated `ArrayInport.Select` optimization is intentionally not part of this PR.

## Semantics and boundaries

`list<T>` and `dict<T>` remain language-level types. The optimization is internal: scalar values can be stored as `[]int64`, `[]string`, etc. (or the analogous typed dictionary); the untyped representation remains `[]Msg` / `map[string]Msg`.

Runtime functions should use `AsList*` / `AsDict*` when the component contract gives a scalar type. `listToMsgs` and `dictToMsgs` deliberately box values only at protocol boundaries that require `Msg` values.

## Performance evidence

Comparison used the same Apple M1 Pro / Go 1.26.5 against clean `main` at `cbdb33a0`; the benchmarked typed-container implementation was `6498232b`. Results are medians of 10 runs with `-benchtime=300ms -benchmem`; dictionary lookup was also repeated with `-cpu=1 -benchtime=500ms`. The follow-up commits only clarify the container API, share invalid-accessor panics, add documentation and tests, and do not change the scalar storage algorithms measured below.

| Scenario | Result |
| --- | --- |
| scalar list iteration, 8–1024 ints | **75–87% faster** |
| construct 512-int list | **82% faster**, 64% fewer bytes, 255 fewer allocations |
| construct 512-int dict | **10.6% faster**, 33.7% fewer bytes, 255 fewer allocations |
| `list_at` / `list_slice` full port path | **3–4% faster**, 1–2 fewer allocations |
| dictionary lookup / `get(dict)` full port path | neutral; no regression observed |

The PR makes a measured claim for scalar-list traversal and for construction/memory footprint of larger scalar containers. It deliberately does **not** claim a standalone dictionary-lookup speedup: channel, tracing, and result boxing dominate the full runtime-function path.

Detailed raw comparison: https://github.com/nevalang/neva/pull/1127#issuecomment-5070041444

## Verification

- current head: `go test ./internal/runtime ./internal/runtime/funcs ./internal/compiler/backend/golang`
- current head: patch-scoped `golangci-lint`
- GitHub Actions were green at `6498232b`; the checks for the current head are running and remain a merge gate.


internal/runtime/program.go
```diff
@@ -257,10 +257,36 @@ func (s SelectedMsg) String() string {
 	return fmt.Sprint(s.OrderedMsg)
 }
 
+func (a *ArrayInport) selectedFromCtx(ctx context.Context, slotIdx int, ordered OrderedMsg) SelectedMsg {
```
emil14: а почему имя такое странное? я бы назвал orderedToSelectedMsg

internal/runtime/program.go
```diff
@@ -303,6 +316,42 @@ func (a ArrayInport) _select(ctx context.Context) ([]SelectedMsg, bool) {
...
+			select {
+			case <-ctx.Done():
+				return SelectedMsg{}, false
+			case ordered := <-a.chans[0]:
+				first = a.selectedFromCtx(ctx, 0, ordered)
+			case ordered := <-a.chans[1]:
+				first = a.selectedFromCtx(ctx, 1, ordered)
+			}
+
+			second, ok := a.tryReceiveSlot(ctx, 0)
```
emil14: оч странная реализация, а если в этот момент уже новое сообщение пришло? мы тогда два раза вычитываемся из одного и того же слота? или это ожидаемо? нужны комментарии

internal/runtime/program.go
emil14: Переименовал в `selectedFromContext` и заодно сделал соседний helper более явным: `tryReceiveSlot` -> `receiveSlotIfReady`. Смысл: название теперь отражает, что мы не просто конвертируем ordered->selected, а одновременно прокидываем через tracer/interceptor контекст слота.

internal/runtime/program.go
emil14: Да, кейс ожидаемый: в двухслотовом fast-path мы читаем первое сообщение блокирующе, затем делаем по одному неблокирующему чтению для поиска конкурента (со второго сообщения). Если конкурент найден — сравниваем `index`, более новое кладем в буфер. Если не найден — сразу отдаем первое. Добавил это явно в комментарий над веткой `len(a.chans)==2`, чтобы алгоритм читался без догадок.

scripts/ci/go-mod-download-with-retry.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
emil14: не надо усложнять, удали это и правки в .github тоже

.github/workflows/lint.yml
```diff
@@ -17,8 +17,10 @@ jobs:
       - uses: actions/checkout@v5
       - uses: actions/setup-go@v6
         with:
-          go-version: '1.26.1'
+          go-version: '1.26.3'
           cache: false
+      - name: download modules (retry)
```
emil14: в этом файле оставлячем только бамп версии го и все

.github/workflows/test.yml
```diff
@@ -14,7 +14,9 @@ jobs:
     - name: setup go
       uses: actions/setup-go@v5
       with:
-        go-version: '1.26.1'
+        go-version: '1.26.3'
```
emil14: тут тоже оставить только бамп

internal/runtime/program.go
```diff
@@ -303,6 +316,46 @@ func (a ArrayInport) _select(ctx context.Context) ([]SelectedMsg, bool) {
 // Select returns oldest available message across all available array inport slots.
 func (a *ArrayInport) Select(ctx context.Context) (SelectedMsg, bool) {
```
emil14, line 297: We need unit tests and benchmarks specifically for this function and TODO comments for all the other public methods in this runtime package, except for those we already covered

internal/compiler/backend/golang/backend.go
```diff
@@ -463,6 +464,56 @@ func (b Backend) getMessageString(msg *ir.Message) (string, error) {
 		}
 		return fmt.Sprintf("runtime.NewUnionMsg(%q, %s)", msg.Union.Tag, payload), nil
 	case ir.MsgTypeList:
+		if len(msg.List) > 0 { //nolint:nestif // TODO(strict-lint phase 1): temporary suppression; remove after strict cleanup.
```
emil14: early return/break not possible?

internal/compiler/backend/golang/backend.go
```diff
@@ -463,6 +464,56 @@ func (b Backend) getMessageString(msg *ir.Message) (string, error) {
...
+			//nolint:exhaustive // Only scalar lists are optimized here; others fallback to generic path.
+			switch msg.List[0].Type {
+			case ir.MsgTypeBool:
+				values := make([]string, len(msg.List))
+				for i := range msg.List {
+					if msg.List[i].Type != ir.MsgTypeBool {
+						break
```
emil14: is there a valid case for this? like `list<any>`? make sure we not trying to cover invalid case, otherwise thats a critical bug. check how type system works for that

internal/compiler/backend/golang/backend.go
```diff
@@ -440,7 +441,7 @@ func (b Backend) buildFuncCalls(
 	return result, nil
 }
 
-//nolint:cyclop,funlen,gocognit,gocyclo // TODO(strict-lint phase 1): temporary suppression; remove after strict cleanup.
+//nolint:cyclop,funlen,gocognit,gocyclo,maintidx // TODO(strict-lint phase 1): temporary suppression; remove after strict cleanup.
 func (b Backend) getMessageString(msg *ir.Message) (string, error) {
```
emil14, line 445: I feel like we could use some unit tests for this func

internal/compiler/backend/golang/backend.go
```diff
@@ -474,6 +525,53 @@ func (b Backend) getMessageString(msg *ir.Message) (string, error) {
 		}
 		return fmt.Sprintf("runtime.NewListMsg([]runtime.Msg{%s})", strings.Join(elements, ", ")), nil
 	case ir.MsgTypeDict:
+		if len(msg.DictOrStruct) > 0 { //nolint:nestif // TODO(strict-lint phase 1): temporary suppression; remove after strict cleanup.
```
emil14: a lot of nolint comments in this PR, is it really that hard to write readable code?

internal/compiler/backend/golang/backend.go
```diff
@@ -474,6 +525,53 @@ func (b Backend) getMessageString(msg *ir.Message) (string, error) {
...
+			var sameType ir.MsgType
+			hasType := false
+			homogeneous := true
+			values := make([]string, 0, len(msg.DictOrStruct))
+			//nolint:gocritic // map iteration by value is acceptable in code generation path.
+			for key, valueMsg := range msg.DictOrStruct {
+				if !hasType {
```
emil14: I wish this code would have more explainning whats going on comments its quite hard to figyre out

internal/runtime/funcs/field.go
```diff
@@ -11,14 +11,19 @@ type structField struct{}
 
 //nolint:varnamelen // TODO(strict-lint phase 1): temporary suppression; remove after strict cleanup.
 func (s structField) Create(io runtime.IO, cfg runtime.Msg) (func(ctx context.Context), error) {
-	path := cfg.List()
-	if len(path) == 0 {
-		return nil, errors.New("field path cannot be empty")
+	var pathStrings []string
+	if typedPath, ok := runtime.AsListStrings(cfg.List()); ok {
+		pathStrings = append(pathStrings, typedPath...)
+	} else {
```
emil14: this comment is related to this specific place AND ALL other places like this in this PR: are you SURE we should cover these "else" cases? For example, is there are valid case for `field` receiving not `list<string>`? remember - runtime TRUSTS compiler. however you should understand the type system to know which guarantees compiler gives. Im afraid we end up with lots of useless safety handling

internal/runtime/message.go
```diff
@@ -201,77 +252,557 @@ func (msg ListMsg) String() string {
...
+
+func dictToString(dict DictMsg) string {
+	b, err := dictMarshalJSON(dict)
+	if err != nil {
+		panic(err)
 	}
+	return string(b)
+}
+
+func AsListMsgs(list ListMsg) ([]Msg, bool) {
```
emil14: wdyt about adding `Must` prefix to all these helper funcs and simply panicking from them if type was guessed incorrectly? will that cover all possible scenarious?

internal/runtime/message.go
```diff
@@ -16,17 +16,19 @@ type OrderedMsg struct {
...
 	Str() string
 	Bytes() []byte
-	List() []Msg
-	Dict() map[string]Msg
+	List() ListMsg
+	Dict() DictMsg
 	Struct() StructMsg
 	Union() UnionMsg
 
+	MarshalJSON() ([]byte, error)
```
emil14: I dont think we need MarshalJSON as part of the Msg interface, why? Can you prove?

internal/runtime/program.go
emil14: Renamed helper to `orderedToSelectedMsg` in `program.go` as requested for explicit intent.

internal/runtime/program.go
emil14: Added explicit inline comment in 2-slot fast-path: double read from same slot is expected and preserves ordering via `OrderedMsg.index` (older returned, newer buffered).

internal/runtime/program.go
emil14: Applied in code: helper naming is now explicit and consistent around SelectedMsg creation.

internal/runtime/program.go
emil14: Kept the 2-slot strategy and documented the exact ordering behavior directly in code to remove ambiguity.

scripts/ci/go-mod-download-with-retry.sh
emil14: Done: removed `scripts/ci/go-mod-download-with-retry.sh` from this PR.

.github/workflows/lint.yml
emil14: Done: reverted this workflow in the branch to keep only the Go version change policy from main.

.github/workflows/test.yml
emil14: Done: reverted this workflow in the branch to keep only the Go version change policy from main.

internal/runtime/program.go
emil14: Added focused unit tests + benchmark for `ArrayInport.Select` in `internal/runtime/program_select_test.go` (single-slot, two-slot ordering, context cancel + hotpath bench).

internal/compiler/backend/golang/backend.go
emil14: Refactored list/dict typed-constructor detection into helpers (`scalarListCtor`, `scalarDictCtor`) to make early fallback explicit and reduce nested flow.

internal/compiler/backend/golang/backend.go
emil14: Confirmed this path intentionally handles valid homogeneous scalar literals only; mixed/non-scalar literals fall back to generic ctor. Added tests in `backend_message_test.go` for both typed and fallback cases.

internal/compiler/backend/golang/backend.go
emil14: Added unit tests for typed constructor selection in `internal/compiler/backend/golang/backend_message_test.go`.

internal/compiler/backend/golang/backend.go
emil14: Reduced several noisy nolint-driven branches by extracting helper functions and simplifying control flow in `getMessageString`.

internal/compiler/backend/golang/backend.go
emil14: Added comments on typed fast-path invariants and generic fallback conditions in backend helpers.

internal/runtime/funcs/field.go
emil14: Agreed and tightened behavior here: `field.Create` now requires `list<string>` config and returns startup error otherwise (no generic fallback path).

internal/runtime/message.go
emil14: Good point. In this iteration I kept `As*` for compatibility with existing callsites, but moved runtime funcs toward typed-first and removed unsafe fallback in `field`. Can do a follow-up to introduce `Must*` APIs consistently.

internal/runtime/message.go
emil14: Accepted: removed `MarshalJSON` from `Msg` interface; marshaling now remains on concrete message/container types and tests were updated accordingly.

internal/runtime/program_select_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	_, ok := in.Select(ctx)
+	if ok {
+		t.Fatal("expected select to stop on canceled context")
+	}
+}
+
+// BenchmarkArrayInportSelectTwoSlots measures the two-slot fast-path in Select:
+// one blocking receive + opportunistic competitor probe + index-based ordering.
+func BenchmarkArrayInportSelectTwoSlots(b *testing.B) {
```
emil14: let's not mix tests and benchmarks in same files, _test.go and _bench_test.go let it be

add this to AGENTS.md

internal/runtime/funcs/utils.go
```diff
@@ -129,6 +129,70 @@ func tryToUnboxIfUnion(msg runtime.Msg) runtime.Msg {
 	return unionMsg.Data()
 }
 
+func listToMsgs(list runtime.ListMsg) []runtime.Msg {
```
emil14, line 130: every newly added entity such as function or type must have doc comments

internal/runtime/funcs/list_at.go
```diff
@@ -64,3 +118,16 @@ func (listAt) Create(io runtime.IO, _ runtime.Msg) (func(ctx context.Context), e
 		}
 	}, nil
 }
+
+//nolint:ireturn // Generic helper returns scalar value type parameter.
+func listItem[T any](data []T, idx int64) (T, bool) {
+	dataLen := int64(len(data))
+	if idx < -dataLen || idx >= dataLen {
```
emil14, line 91: explain all invariants of this function in doc comment and have unit tests for it

internal/runtime/funcs/field.go
```diff
@@ -11,14 +11,14 @@ type structField struct{}
 
 //nolint:varnamelen // TODO(strict-lint phase 1): temporary suppression; remove after strict cleanup.
 func (s structField) Create(io runtime.IO, cfg runtime.Msg) (func(ctx context.Context), error) {
-	path := cfg.List()
-	if len(path) == 0 {
-		return nil, errors.New("field path cannot be empty")
+	typedPath, ok := runtime.AsListStrings(cfg.List())
```
emil14, line 14: so what did we decide about turning these funcs into MustXXX?

internal/runtime/funcs/get.go
```diff
@@ -55,16 +55,64 @@ func (g getDictValue) Create(io runtime.IO, _ runtime.Msg) (func(ctx context.Con
 				return
 			}
 
-			valueMsg, ok := dictMsg.Dict()[keyMsg.Str()]
-			if !ok {
-				if !errOut.Send(ctx, errFromString("Key not found in dictionary")) {
+			key := keyMsg.Str()
```
emil14: видимо этот компонент тоже должен быть перегруженый

internal/runtime/funcs/list_at.go
```diff
@@ -42,7 +42,61 @@ func (listAt) Create(io runtime.IO, _ runtime.Msg) (func(ctx context.Context), e
 			}
 
 			idx := idxMsg.Int()
-			data := dataMsg.List()
+			list := dataMsg.List()
+			if data, ok := runtime.AsListInts(list); ok {
```
emil14: 1. лучше бы тут был свич
2. но вообще видимо этот компонент тоже должен быть перегруженый

я тут пока писал подумал, мы таким макаром придем к тму что у нас сколько компонентов перегруженные будут?

internal/runtime/program.go
```diff
@@ -257,10 +257,36 @@ func (s SelectedMsg) String() string {
...
+func (a *ArrayInport) orderedToSelectedMsg(ctx context.Context, slotIdx int, ordered OrderedMsg) SelectedMsg {
+	index := Uint8Index(slotIdx)
+	slotAddr := PortSlotAddr{
+		PortAddr: PortAddr{
+			Path: a.addr.Path,
+			Port: a.addr.Port,
+		},
+		Index: &index,
+	}
+	a.tracer.recordReceived(slotAddr, ordered)
```
emil14: имя функции намекает на чистую функцию - трансформер, а по факту тут сайд эффекты. наверно лучше вынести трейсер и интерсептор в вызывающую сторону

internal/runtime/message.go
emil14: аргументируй почему решил не делать как я предложил

internal/runtime/message.go
```diff
@@ -201,77 +251,557 @@ func (msg ListMsg) String() string {
...
+	case floatDictMsg:
+		return typed.MarshalJSON()
+	case stringDictMsg:
+		return typed.MarshalJSON()
+	default:
+		panic("unexpected dict implementation")
+	}
+}
+
+func dictToString(dict DictMsg) string {
```
emil14, line 743: а почему не просто String() на дикте?

internal/runtime/message.go
```diff
@@ -201,77 +251,557 @@ func (msg ListMsg) String() string {
...
+
+func listToString(list ListMsg) string {
+	b, err := listMarshalJSON(list)
+	if err != nil {
+		panic(err)
+	}
+	return string(b)
+}
+
+func dictMarshalJSON(dict DictMsg) ([]byte, error) {
```
emil14, line 726: надо иметь в виду, что вызово вот таких методов где мы type assertion занимаемся должно быть как можно меньше в рантайме, тут хорошо бы по максимуму ожидать что мы знаем тип наверняка, не факт что это всегда возможно, но надо к этому стремиться, для этого надо понимать type system в компиляторе

internal/runtime/message.go
```diff
@@ -201,77 +251,557 @@ func (msg ListMsg) String() string {
...
+		return false
 	}
 	return true
 }
 
-func NewDictMsg(d map[string]Msg) DictMsg {
-	return DictMsg{
-		internalMsg: internalMsg{},
-		v:           d,
+func asGenericDict(dict DictMsg) map[string]Msg {
```
emil14: такого должно быть как можно меньше

internal/runtime/program_select_test.go
emil14: Сделал: вынес benchmark из  в отдельный . Также добавил правило в : unit tests и benchmarks держим в отдельных файлах.

internal/runtime/funcs/utils.go
emil14: Принял. Добавил doc comments для новых helpers в  (, , ).

internal/runtime/funcs/list_at.go
emil14: Добавил подробный doc comment с инвариантами над  и добавил unit tests в  (typed fast-path, generic fallback, out-of-bounds, negative index).

internal/runtime/funcs/field.go
emil14: По MustXXX: в этом PR не переводил API массово в , чтобы не расширять scope. Текущая итерация держит typed-first extractors и явные fail-fast точки. Если ок, следующим шагом вынесу отдельным PR точечный переход на  там, где контракт строго гарантирован компилятором.

internal/runtime/funcs/get.go
emil14: Согласен по направлению. В этой итерации усилил typed dict-path в , но полную перегрузку компонента предлагаю отдельным шагом (чтобы не раздувать этот PR и не смешивать с текущим fix-batch).

internal/runtime/funcs/list_at.go
emil14: Сделал  через явный helper-path с -веткой typed-first. По вопросу про масштаб перегрузок: предлагаю фиксировать policy отдельным RFC/issue и двигаться по hotpath-компонентам в порядке профита (а не перегружать всё подряд сразу).

internal/runtime/program.go
emil14: Исправил по смыслу: переименовал helper в , чтобы имя явно отражало side effects (trace/interceptor). Вынесение side effects наружу сейчас не делал, чтобы не разломать горячий путь в этом PR.

internal/runtime/program_select_test.go
emil14: Сделал: вынес benchmark из program_select_test.go в отдельный program_select_bench_test.go. Также добавил правило в AGENTS.md: unit tests и benchmarks держим в отдельных файлах.

internal/runtime/funcs/utils.go
emil14: Принял. Добавил doc comments для новых helpers в internal/runtime/funcs/utils.go (tryToUnboxIfUnion, listToMsgs, dictToMsgs).

internal/runtime/funcs/list_at.go
emil14: Добавил подробный doc comment с инвариантами над listAt.Create и добавил unit tests в internal/runtime/funcs/list_at_test.go (typed fast-path, generic fallback, out-of-bounds, negative index).

internal/runtime/funcs/field.go
emil14: По MustXXX: в этом PR не переводил API массово в Must*, чтобы не расширять scope. Текущая итерация держит typed-first extractors и явные fail-fast точки. Если ок, следующим шагом вынесу отдельным PR точечный переход на Must* там, где контракт строго гарантирован компилятором.

internal/runtime/funcs/get.go
emil14: Согласен по направлению. В этой итерации усилил typed dict-path в get, но полную перегрузку компонента предлагаю отдельным шагом (чтобы не раздувать этот PR и не смешивать с текущим fix-batch).

internal/runtime/funcs/list_at.go
emil14: Сделал list_at через явный helper-path с switch-веткой typed-first. По вопросу про масштаб перегрузок: предлагаю фиксировать policy отдельным RFC/issue и двигаться по hotpath-компонентам в порядке профита (а не перегружать всё подряд сразу).

internal/runtime/program.go
emil14: Исправил по смыслу: переименовал helper в recordAndSelectMsg, чтобы имя явно отражало side effects (trace/interceptor). Вынесение side effects наружу сейчас не делал, чтобы не разломать горячий путь в этом PR.

internal/runtime/message.go
emil14: Хороший пойнт. Сейчас оставил dictValueMsg.String() через общий formatter и добавил комментарий почему: DictMsg контракт намеренно не требует Stringer; форматирование централизовано в message-level helper.

internal/runtime/message.go
emil14: Согласен полностью. В этой волне сократил часть мест, где уходим в generic path, и в hotpath оставил typed-first извлечения. Дальше план: продолжать убирать runtime type-assertions из горячих участков, где есть гарантии от компилятора.

internal/runtime/message.go
emil14: Принял, согласен. В текущем батче дополнительно сдвинул код в сторону typed-first и минимизации generic/type-assert путей на горячих участках.

emil14:
## Benchmark comparison: typed containers vs current main

**Method.** Same Apple M1 Pro / Go 1.26.5; clean `main` at `cbdb33a0` versus merged candidate `6498232b`. Each result is the median of 10 runs with `-benchtime=300ms -benchmem`. The runtime-function baseline uses the old `[]Msg` / `map[string]Msg` representation for the same port protocol. Dict-lookup microbenchmarks were additionally repeated with `-cpu=1 -benchtime=500ms` to rule out scheduler noise.

### What is measurably faster

| Scenario | main | PR | change |
| --- | ---: | ---: | ---: |
| List iteration, 8 ints | 19.89 ns/op | 4.959 ns/op | **-75.1%** |
| List iteration, 64 ints | 191.5 ns/op | 35.43 ns/op | **-81.5%** |
| List iteration, 512 ints | 1458 ns/op | 223.8 ns/op | **-84.7%** |
| List iteration, 1024 ints | 3309 ns/op | 435.9 ns/op | **-86.8%** |
| Construct list, 64 ints | 394.4 ns, 1176 B, 2 allocs | 146.4 ns, 552 B, 3 allocs | **-62.9%; -53.1% bytes** |
| Construct list, 512 ints | 4409 ns, 11544 B, 258 allocs | 794.0 ns, 4136 B, 3 allocs | **-82.0%; -64.2% bytes; -255 allocs** |
| Construct dict, 512 ints | 33822 ns, 46704 B, 1184 allocs | 30238 ns, 30972 B, 929 allocs | **-10.6%; -33.7% bytes; -255 allocs** |
| `list_at` full port path | 2951 ns, 830 B, 12 allocs | 2863 ns, 805 B, 11 allocs | **-3.0%; -1 alloc** |
| `list_slice` full port path | 4435 ns, 1136 B, 27 allocs | 4259 ns, 1083 B, 25 allocs | **-4.0%; -2 allocs** |

### Neutral, deliberately not claimed as wins

| Scenario | main | PR | conclusion |
| --- | ---: | ---: | --- |
| Dict hot lookup, 16 ints | 12.80 ns/op | 12.80 ns/op | neutral |
| Dict hot lookup, 128 ints | 12.78 ns/op | 12.70 ns/op | neutral |
| Dict mixed lookup, 16 ints | 199.9 ns/op | 200.6 ns/op | neutral |
| Dict mixed lookup, 128 ints | 1565 ns/op | 1564 ns/op | neutral |
| `get(dict)` full port path | 2998 ns, 868 B, 15 allocs | 2995 ns, 868 B, 15 allocs | neutral |

Conclusion: the PR has a strong, repeatable performance case for scalar **list** traversal and for construction/memory footprint of larger scalar list/dict values. It does not establish a standalone speedup for map lookup, and the full runtime-function benefits are modest because channels, tracing, and result boxing dominate there.

Validation at `6498232b`:
```
GOCACHE=/private/tmp/neva-1127-pr-gocache go test ./internal/runtime ./internal/runtime/funcs ./internal/compiler/backend/golang
```

The new PR head is based on current `main`; CI has been restarted and must be green before merge.

internal/runtime/message.go
```diff
@@ -186,13 +194,55 @@ func NewBytesMsg(v []byte) BytesMsg {
...
+func (msg listValueMsg) MarshalJSON() ([]byte, error) {
+	return listMarshalJSON(msg.v)
+}
+
+func (msg listValueMsg) Equal(other Msg) bool {
+	otherList, ok := other.(listValueMsg)
+	return ok && msg.v.Equal(otherList.v)
+}
+
+type genericListMsg struct{ v []Msg }
```
emil14: let's have internalListMsg struct that implements all methods by panicking and then we embed it into every listMsg struct instaed of duplicating panics mehtods

internal/runtime/message.go
emil14: answer, do not change the code

what do you think if instead of `runtime.Msg.Equal(Msg)` we would have `runtime.Equal(Msg, Msg)`?

wouldn't it make the `Msg` interace more clear?

why I'm asking - looking at the code now made me thinking: "there's a lot of possible operations that could be done with messages, and it's not clear to me how exactly the concernts must be separated between the runtime package and runtime funcs package, for example logic of getting list element by index is part of the runtime function, but that's working with memory layout which is kinda defined by runtime/message.go file"

I might be wrong so I want you to answer from the perspective of a clean code architecht, best practices specialist

internal/runtime/message.go
```diff
@@ -186,13 +194,55 @@ func NewBytesMsg(v []byte) BytesMsg {
...
 //nolint:godoclint // TODO(strict-lint phase 1): temporary suppression; remove after strict cleanup.
-type ListMsg struct {
+type ListMsg interface {
+	Msgs() []Msg
+	Bools() []bool
+	Ints() []int64
+	Floats() []float64
+	Strings() []string
+	Len() int
+	Equal(ListMsg) bool
```
emil14: answer, do not change the code

this is continuation of the question https://github.com/nevalang/neva/pull/1127/changes#r3650677457

It bothers me a bit that two things are combined in these interfaces: 1) accessing the proper memory field 2) some kind of operation on the message container (len + equal in this case), e.g. we also have things like "get by idx" or "slice", why those are runtime funcs part and not present here?

in a world where runtime functions just call runtime APIs for such things, is this better picture or worse and why?

internal/runtime/message.go
```diff
@@ -186,13 +194,55 @@ func NewBytesMsg(v []byte) BytesMsg {
...
+}
+
+//nolint:godoclint // TODO(strict-lint phase 1): temporary suppression; remove after strict cleanup.
+type listValueMsg struct {
```
emil14, line 185: am I right that this type only exist to avoid circular dependency between Msg and List/DictMsg types?

e.g. can't all List implementations embed internalMsg themselves so things like NewStringList() return just stringListMsg instead of this intermediate wrapper?

also in one of the comments above I ask to introduce internalListMsg (and internalDictMsg therefore) structs that implements panicking methods, can those embed internalMsg message so we don't even have to embed internalMsg in the list msg implementations?

or that would cause problems like internalMsg.List() panics, while stringListMsg does not have List method implemented, yet being a list actually

more and foremost listValueMsg even if needed must have a doc-comment explaining what is it and what for

internal/runtime/message.go
```diff
@@ -201,77 +251,560 @@ func (msg ListMsg) String() string {
...
-	if !ok {
-		return false
+func (msg genericListMsg) MarshalJSON() ([]byte, error) {
+	return json.Marshal(msg.v)
+}
+func (msg genericListMsg) Equal(other ListMsg) bool {
+	return listEqualGeneric(msg.v, other)
+}
+
+func (boolListMsg) Msgs() []Msg        { panic("unexpected Msgs method call on bool list message") }
```
emil14: I think Msgs() is bad name, would be better `Untyped()`

internal/runtime/message.go
```diff
@@ -201,77 +251,560 @@ func (msg ListMsg) String() string {
...
+//
+//nolint:godoclint // TODO(strict-lint phase 1): temporary suppression; remove after strict cleanup.
+type DictMsg interface {
+	Msgs() map[string]Msg
+	Bools() map[string]bool
+	Ints() map[string]int64
+	Floats() map[string]float64
+	Strings() map[string]string
+	Len() int
+	Equal(DictMsg) bool
```
emil14: Similarly to my other questions about perfect runtime API and runtime functions API and separation of concerns - there's a lot of possible operations on top of lists and dicts, not just len and equal, also possible to get by index/key, merge, add, maybe something else. One way (not sure perfect tho) is to see it as "if its expressed in Go/Python as syntax, we should have it in runtime package). Note that in Go stuff like `len` and `make` is not syntax but part of the language too.

internal/runtime/message.go
```diff
@@ -201,77 +251,560 @@ func (msg ListMsg) String() string {
...
+func NewDictFloatMsg(d map[string]float64) Msg {
+	return dictValueMsg{internalMsg: internalMsg{}, v: floatDictMsg{v: d}}
+}
+
+//nolint:ireturn // Msg contract type.
+func NewDictStringMsg(d map[string]string) Msg {
+	return dictValueMsg{internalMsg: internalMsg{}, v: stringDictMsg{v: d}}
+}
+
+func dictEqual(left DictMsg, right DictMsg) bool {
```
emil14, line 502: every new entity in go code MUST have doc comments

internal/runtime/message.go
```diff
@@ -201,77 +251,560 @@ func (msg ListMsg) String() string {
...
+	b, err := msg.MarshalJSON()
+	if err != nil {
+		panic(err)
+	}
+	return string(b)
+}
+
+func listMarshalJSON(list ListMsg) ([]byte, error) {
+	switch typed := list.(type) {
+	case genericListMsg:
```
emil14: let's call it untyped(List|Dict)Msg insted of generic to avoid confusion with type system parameters/arguments

internal/runtime/message.go
```diff
@@ -201,77 +251,560 @@ func (msg ListMsg) String() string {
...
+
+func dictToString(dict DictMsg) string {
+	b, err := dictMarshalJSON(dict)
+	if err != nil {
+		panic(err)
 	}
+	return string(b)
+}
+
+func AsListMsgs(list ListMsg) ([]Msg, bool) {
```
emil14: we have a LOT of APIs about lists and dicts and its genuinely hard to tell if this design is the best one...

internal/runtime/funcs/dict_get_by_key.go
```diff
@@ -9,7 +9,7 @@ import (
 
 type getDictValue struct{}
```
emil14: lets rename file to dict_get_by_key.go, struct to dictGetByKey, in the neva layer keep as is just Get for now

internal/runtime/funcs/utils.go
```diff
@@ -122,6 +125,74 @@ func tryToUnboxIfUnion(msg runtime.Msg) runtime.Msg {
 	return unionMsg.Data()
 }
 
+// listToMsgs converts any supported typed list view to generic []runtime.Msg.
+// Typed scalar paths avoid panicking Msgs() calls on typed list implementations.
+func listToMsgs(list runtime.ListMsg) []runtime.Msg {
```
emil14, line 130: I feel like this kind of logic should be better part of the runtime package

internal/runtime/message.go
emil14: Done in 81b2ee11. `internalListMsg` and `internalDictMsg` now own every invalid accessor and panic consistently; each concrete storage type embeds the relevant internal base and implements only its valid accessor, `Len`, `Equal`, and JSON formatting. I also added tests for valid untyped/typed access and invalid accessor panics.

internal/runtime/message.go
emil14: I would keep `Msg.Equal(Msg)` rather than introduce `runtime.Equal(Msg, Msg)`. Equality is representation-polymorphic: its efficient and correct implementation is owned by the concrete message type, and method dispatch makes that explicit. A package-level function would either immediately type-switch over every concrete message (duplicating the same closed-world knowledge) or just call the method, adding an indirection without making ownership clearer.\n\nThe boundary I would use is: runtime owns data representation, representation-preserving primitives and cross-representation invariants (`Equal`, `Len`, typed accessors, JSON); runtime functions own graph-component semantics — port protocol, pairing inputs, error messages, bounds and missing-key behavior, output construction and causes. `list_at` is not merely an index operation: it defines the component error/protocol behavior. We should keep it there unless we deliberately introduce a reusable, error-policy-free runtime primitive.

internal/runtime/message.go
emil14: Yes, the two concerns need a firmer boundary. `ListMsg`/`DictMsg` should stay a storage-view API: which representation is present, how long it is, and value equality across representations. They should not grow into a grab-bag of every operation supported by the language.\n\n`get by index`, `slice`, and `get by key` currently belong to runtime functions because they carry component-level semantics: input/output ports, ordering causes, and user-visible errors for bounds/missing keys. Moving them verbatim into `runtime` would force runtime to choose those semantics and make the interface grow with every std component. A better future extraction, if profiling shows reuse, is small policy-free primitives such as `ListValueAt(list, index) (Msg, bool)`; the runtime function would still decide what a false result means and how to send it. This PR intentionally stops at typed storage access, not that broader API redesign.

internal/runtime/message.go
emil14: `listValueMsg` is not required to break a Go circular dependency: Go can express `Msg` ↔ `ListMsg` references. It is an intentional adapter between two layers. `listValueMsg` is the `Msg` payload; `untypedListMsg`/scalar list structs are storage implementations that only satisfy `ListMsg`. If each storage implementation embedded `internalMsg`, it would need to override all unrelated `Msg` methods and make every storage view a full message. That increases coupling and leaves more accidental surface.\n\nThe wrapper therefore keeps the conversion explicit and makes `Msg.List()` the only entry into the storage view. I added the missing doc comment explaining this role. Separately, `internalListMsg`/`internalDictMsg` now deduplicate the invalid storage accessors as proposed in the other thread.

internal/runtime/message.go
emil14: Done in 81b2ee11: `Msgs()` is now `Untyped()` throughout the list/dict storage API. The name says exactly what the accessor returns and avoids implying that scalar typed containers can produce `[]Msg` without boxing.

internal/runtime/message.go
emil14: I would not use “everything expressible as Go/Python syntax” as the split: that would make `runtime` mirror an open-ended container library and blur language-component semantics with storage. The useful split is lower-level: runtime owns representation and operations whose contract is independent of a Neva component; runtime functions own public graph behavior and error/protocol policy.\n\nSo `Len` and `Equal` belong to runtime because they are representation invariants. `slice`, lookup, merge and append should become runtime primitives only when multiple components need the same policy-free operation or profiling requires it. Until then, each added primitive is an API cost, and the runtime function remains the narrowest owner of the component contract.

internal/runtime/message.go
emil14: Done in 81b2ee11. I added doc comments for the new exported container contracts, constructors and representation extractors, plus the new internal adapter/storage types. The comments state which accessor is valid, where the panic invariant applies, and why the value wrappers exist.

internal/runtime/message.go
emil14: Done in 81b2ee11: `genericListMsg`/`genericDictMsg` and related helper names are now `untypedListMsg`/`untypedDictMsg`. This separates runtime representation terminology from language-level generics.

internal/runtime/message.go
emil14: Agreed: the interface is currently a deliberately small storage API, not a claim that this is the final general container API. The concrete conclusion I am applying here is to keep it limited to representation access, `Len`, and `Equal`; adding lookup/slice/merge/append requires an explicit policy decision, because those operations bring error and allocation semantics. I would capture the broader runtime-vs-runtime-functions API design as a dedicated follow-up rather than let this performance PR choose it implicitly.

internal/runtime/funcs/dict_get_by_key.go
emil14: Done in 81b2ee11. The implementation is now `dictGetByKey` in `dict_get_by_key.go`; the registry key and Neva-facing component remain `get_dict_value` / `Get` as requested.

internal/runtime/funcs/utils.go
emil14: I agree with the ownership direction, but I am not moving this blindly in this PR. `listToMsgs`/`dictToMsgs` are representation conversion: they box scalar storage into `Msg` values, so they conceptually belong next to the container representation. Moving them out of `funcs` requires choosing a new exported runtime API (`runtime.ListToMsgs` / `runtime.DictToMsgs` or a better name) and documenting its allocation guarantee; that becomes a stable internal package surface.\n\nMy proposed next step is to move them to `runtime` after we agree on that API name and boundary. I do not want to introduce an exported conversion operation as an incidental consequence of this performance PR, especially since typed-preserving call sites should avoid it. For now the helpers are explicitly documented as the boxing boundary and isolated to the only callers that need boxed values.

Issue #761, opened 2024-11-12
Implicit `any` in node expressions causes problems


emil14:
# Problem

Let's first look at this `Main` component and specifically on how `For` is used

```neva
def Main(start any) (stop any) {
	For{Next2Lines}, Wait
	---
	:start -> 99..-1 -> for -> wait -> :stop
}
```

As we can see it used without type-arguments even though it has one type-parameter

```neva
pub def For<T>(data stream<T>) (res stream<T>) {
    ...
    handler IForEachHandler<T>
    lock1 Lock<stream<T>>
    ...
    ---
    ...
    lock1 -> handler // FIXME compiler must not allow this connection
    ...
}

pub interface IForEachHandler<T>(data T) (sig any)
```

This is because compiler allows this and implicitly uses `any` as a default type-argument. This is however lead to a non-obvious bug that `lock1 -> handler` connection is allowed even though `T` and `stream<T>` cannot be same thing... Right? Well, turns out they can if `T` is `any`!

```
stream<any> :> any
```

# Proposal

Get rid of this feature and enforce explicit type-arguments. Implicit `any` is basically type inference of smoker. Unless we heave real type-inference we must not make assumptions

PR #1132, opened 2026-05-26
analyzer: remove implicit any default for generic node instantiations

emil14:
## Summary
- remove implicit `any` fallback in analyzer for nodes with one type parameter
- require explicit type args at callsites that relied on fallback
- update std/examples callsites to preserve current behavior with explicit `<any>` where appropriate

## Why
Fixes implicit typing behavior tracked in #761 and makes node typing explicit/consistent.

## Validation
- `go test ./internal/compiler/analyzer/...`
- `go test ./examples/...`


internal/compiler/analyzer/nodes.go
```diff
@@ -208,18 +208,6 @@ func (a Analyzer) analyzeNode( //nolint:cyclop,funlen,gocognit,lll,maintidx // T
 		}
 	}
 
-	// default any
```
emil14, line 211: we must have unit tests for this

PR #1144, opened 2026-07-15
Validate union literal payloads

emil14:
Extracts general union-literal payload inference and its regression test from #1042. This remains a separate compiler/type-system follow-up; it is temporarily based on #1042 because that branch contains the surrounding union-literal work. After #1042 merges, retarget this PR to main.


e2e/union_literal_payload_type_mismatch/main/main.neva
```diff
@@ -0,0 +1,9 @@
+type U union {
+	Int int
+}
+
+def Main(start any) (stop any) {
+	pass Pass<U>
+	---
+	:start -> U::Int('bad') -> pass -> :stop
```
emil14, line 8: Do we have support for union literals with payload? Didn't we remove it when we were removing "wrapped senders"?

e2e/union_literal_payload_type_mismatch/neva.yml
```diff
@@ -0,0 +1 @@
+neva: 0.38.0
```
emil14: aren't we on 0.39?

internal/compiler/analyzer/network.go
```diff
@@ -879,6 +879,18 @@ func (a Analyzer) getConstSenderType(
 		}
 	}
 
+	if constSender.Value.Message.Union != nil {
```
emil14, line 882: shouldn't we have a unit test for analyzer code?

internal/compiler/analyzer/union_logic.go
```diff
@@ -6,13 +6,122 @@ import (
...
 	src "github.com/nevalang/neva/pkg/ast"
+	"github.com/nevalang/neva/pkg/core"
 )
 
 type unionActiveTagInfo struct {
 	tag         string
 	tagTypeExpr ts.Expr
 }
 
+func (a Analyzer) inferUnionLiteralSenderType(
```
emil14, line 18: 1. shouldn't we have a unit test for analyzer code? 
2. we MUST have doc-comments for EACH function we create

e2e/union_literal_payload_type_mismatch/main/main.neva
emil14: Yes. Payload-bearing tagged-union literals are supported as static message literals: Day::Friday(value). The language docs explicitly list this form, the parser accepts it, and the Go backend emits runtime.NewUnionMsg(tag, payload).

What we removed with wrapped senders was the distinct dynamic form where the payload came from another network sender; that obscured the graph's data flow. This test intentionally uses a static literal and verifies that U::Int('bad') is rejected before a generic receiver can erase the selected member's payload type. The explicit Union<T> component remains the way to wrap dynamic data.

e2e/union_literal_payload_type_mismatch/neva.yml
emil14: Updated the fixture manifest to neva: 0.39.0 in af4b56df.

internal/compiler/analyzer/network.go
emil14: Added a focused analyzer unit test in internal/compiler/analyzer/union_logic_test.go. It calls the payload-validation path directly and asserts that U::Int('bad') is rejected for Int int. The existing e2e test remains as the end-to-end regression covering the parser, analyzer, and CLI.

internal/compiler/analyzer/union_logic.go
emil14: Done in af4b56df: added the analyzer unit test and doc-comments for the two newly introduced functions, inferUnionLiteralSenderType and messageLiteralType. The other helper functions introduced around this path already have doc-comments.

Issue #1148, opened 2026-07-25
runtime: separate message storage from value operations


emil14:
## Context

PR #1127 introduces typed scalar list/dict storage. During review we identified a broader architecture task that must not block this performance PR: runtime functions should orchestrate ports and graph errors, while `internal/runtime` owns pure operations on message values and their storage representations.

## Goal

Make the boundary explicit and keep it documented:

- `internal/runtime/messages`: immutable message values, storage representations, pure value operations, and allocation/boxing behavior.
- `internal/runtime`: ports, ordering envelopes, tracing, program execution, and other transport/concurrency concerns.
- `internal/runtime/funcs`: receive/send, concurrent input protocol, causes/tracing, and mapping pure-operation failures to public Neva errors.

No port I/O, context, tracer, or user-facing error text belongs in message operations.

## Target API

Keep `ListMsg` and `DictMsg` as storage views. Their scalar/untyped accessors remain useful representation access:

- `Untyped() []Msg` / `Untyped() map[string]Msg` for already boxed containers.
- typed accessors or equivalent `AsListInts`, `AsDictStrings`, etc. for unboxed scalar storage.

Move operations out of those interfaces and into package-level `messages` APIs. The exact initial set should be:

```go
func Equal(left, right Msg) bool
func ListLen(list ListMsg) int
func ListAt(list ListMsg, index int64) (Msg, bool)
func ListSlice(list ListMsg, from, to int64) (Msg, bool)
func ListAppend(list ListMsg, value Msg) (Msg, bool)
func ListPrepend(list ListMsg, value Msg) (Msg, bool)
func ListConcat(left, right ListMsg) Msg
func DictGet(dict DictMsg, key string) (Msg, bool)
func ListToMsgs(list ListMsg) []Msg
func DictToMsgs(dict DictMsg) map[string]Msg
```

`ListToMsgs` and `DictToMsgs` deliberately allocate and box typed scalar values; document that contract. Operations returning a list should preserve typed storage when their inputs permit it. `bool` reports a representation-independent failure such as an invalid index or incompatible append; runtime functions decide the public error message.

Do not add all conceivable container operations preemptively. Add a runtime operation when it is pure and either is shared by multiple runtime functions or owns a representation-sensitive algorithm.

## Package boundary and first PR

[...]

PR #1154, opened 2026-07-26
runtime: move pure value operations to messages

emil14:
Implements the remaining pure value-operation boundary from #1148.

- co-locates each message representation and its pure operations in a single source file: list, bool, int, float, string, dict, struct, and so on;
- moves list access, slicing, immutable append, and immutable prepend to runtime/messages while preserving typed storage;
- renames the public list operation from Push to Append and adds Prepend;
- moves scalar arithmetic, comparison, conversion, parsing, formatting, string transforms, regexp submatch, struct traversal and template formatting to runtime/messages;
- leaves runtime functions responsible only for ports, ordering/causes, external I/O, state, and public-error routing;
- adds direct value tests and direct ListSlice/ListPrepend benchmarks.

Validation: go test ./..., focused lint/tests, selected e2e coverage, and direct message benchmarks.


internal/runtime/messages/list_operations.go
```diff
@@ -0,0 +1,128 @@
+package messages
```
emil14: move all related to list stuff to list.go, same for bool, string, etc. do not create more files

internal/runtime/funcs/image.go
```diff
@@ -19,10 +19,10 @@ type rgbaMsg struct {
 
 func (c *rgbaMsg) decode(msg messages.Msg) {
 	m := msg.Struct()
-	c.r = m.Get("r").Int()
-	c.g = m.Get("g").Int()
-	c.b = m.Get("b").Int()
-	c.a = m.Get("a").Int()
+	c.r = messages.StructGet(m, "r").Int()
```
emil14: StructGet -> StructGetField

internal/runtime/funcs/list_append.go
```diff
@@ -35,13 +35,9 @@ func (p listPush) Create(io runtime.IO, _ messages.Msg) (func(ctx context.Contex
...
 			}
 
-			lstCopy := messages.ListToMessageSlice(lstMsg.List())
-
 			if !resOut.Send(
 				ctx,
-				messages.NewListMsg(
-					append(lstCopy, dataMsg.Msg),
-				),
+				messages.ListAppend(lstMsg.List(), dataMsg.Msg),
```
emil14, line 40: let's rename list_push to list_append (file) and listAppend (type), also let's rename neva component too (I dont remember its name)

internal/runtime/funcs/parse_bool.go
```diff
@@ -42,7 +41,7 @@ func (parseBool) Create(
 				return
 			}
 
-			parsed, parseErr := strconv.ParseBool(data.Str())
+			parsed, parseErr := messages.BoolFromString(data.Msg)
```
emil14, line 44: what do you think about the fact that we do have general Equal() function in messages.Equal(left, right) and same for Match, but other operations are all only specific? is it good? I was thinking about having things like e.g. "And" or "IsGreater" the same as `message.And(left, right) bool` and inside there would be matching. is it bad idea? what's the difference (if any) between Equal+Match and other stuff?

Only answer, do not change the code based on this comment

internal/runtime/messages/list_operations.go
```diff
@@ -0,0 +1,128 @@
...
+		panic("unexpected list implementation")
+	}
+}
+
+// ListAppend returns a new list with value appended. It preserves typed scalar
+// storage when value has the matching scalar representation; otherwise it
+// returns an untyped list.
+//
+//nolint:ireturn // Msg is the value-layer contract.
+func ListAppend(list ListMsg, value Msg) Msg {
```
emil14: let's add list prepend too, with runtime functon, unit tests and benchmarks

internal/runtime/messages/list_operations.go
emil14: Done in e3c8e47c: removed the separate *_operations.go files. Each representation now keeps its pure operations in its own file: list.go, bool.go, int.go, float.go, string.go, etc.

internal/runtime/funcs/image.go
emil14: Done in e3c8e47c: renamed the package operation and all production call sites to messages.StructGetField.

internal/runtime/funcs/list_append.go
emil14: Done in e3c8e47c: renamed the runtime implementation/file to listAppend/list_append.go, the registry key to list_append, and the public Neva component from lists.Push to lists.Append.

internal/runtime/funcs/parse_bool.go
emil14: Equal and Match are genuinely polymorphic operations: list<any>, dict<any>, structs and unions can compare or match recursively, and different runtime kinds are a normal result (Equal returns false). And/IsGreater are different: they are only meaningful for particular operand families and the compiler already guarantees that family. A generic messages.And(left, right) would therefore need its own runtime type dispatcher plus a new error/panic policy for unsupported pairs, duplicate the static type system, and obscure the concrete fast path. Keeping BoolAnd, IntIsGreater, FloatIsGreater, etc. expresses their precondition and yields a concrete result. The useful boundary is not generic versus specific; it is whether the language operation has a total cross-kind semantic. Equal/Match do; arithmetic, ordering, and boolean algebra do not.

internal/runtime/messages/list_operations.go
emil14: Done in e3c8e47c: added messages.ListPrepend, the list_prepend runtime function and public lists.Prepend component, plus direct message and runtime-function unit tests and BenchmarkListPrepend.

internal/runtime/messages/int.go
```diff
@@ -28,3 +28,66 @@ func equalIntValue(left int64, right Msg) bool {
...
+func IntIncrement(value Msg) IntMsg { return NewIntMsg(value.Int() + 1) }
+
+// IntDecrement returns value minus one.
+func IntDecrement(value Msg) IntMsg { return NewIntMsg(value.Int() - 1) }
+
+// IntPower returns base raised to exponent.
+func IntPower(base, exponent Msg) IntMsg {
+	result := int64(1)
+	for range exponent.Int() {
+		result *= base.Int()
```
emil14: are you sure this is the most efficient way to perform this? doesn't Golang provide stdlib's mechanisms to call lower-level code to do something like this? I'm worried about not only this function but the whole runtime package in general, messages and funcs both

internal/runtime/messages/list.go
```diff
@@ -393,6 +393,172 @@ func ListToMessageSlice(list ListMsg) []Msg {
...
 }
 
+// ListAt returns the item at index. Negative indexes count from the end.
+// The boolean is false when index is outside the list bounds.
+//
+//nolint:ireturn // Msg is the value-layer contract.
+func ListAt(list ListMsg, index int64) (Msg, bool) {
+	switch typed := list.(type) {
+	case untypedListMsg:
+		value, found := listItem(typed.v, index)
```
emil14: call helper function `listAt` so that it's clear that it's just a private sub-method of the ListAt function, follow same pattern for everything in this PR

internal/runtime/messages/int.go
emil14: Good catch. Go has no standard signed-int64 exponentiation primitive: math.Pow works in float64 and would change precision, overflow, and negative-base semantics; math/bits provides bit operations but no power. In a5d5677c I replaced the linear loop with integer exponentiation by squaring, preserving the existing negative-exponent result (1) and integer overflow behavior while reducing the work from O(exponent) to O(log exponent). The runtime uses ordinary typed Go arithmetic deliberately: the compiler emits native instructions; we should improve algorithms where a general-purpose implementation is actually on a hot path rather than introduce unsafe or float-based substitutes.

internal/runtime/messages/list.go
emil14: Done in a5d5677c: renamed the private list helpers to the public-operation pattern: listAt, listSlice, listAppend, listPrepend, listSliceBounds, and listSliceIndex. String indexing/slicing uses the same named primitives where appropriate.

PR #1157, opened 2026-07-26
runtime: preserve typed containers from streams

### LINKED ISSUE #908: Redesign of `stream`

### LINKED ISSUE #1156: runtime: preserve typed containers across stream materialization
## Context

This is the explicit follow-up to #1148 (closed by #1155). That issue moved existing pure value operations into `internal/runtime/messages`, but its final audit did not cover every runtime path that materializes a list or dictionary.

The current stream protocol is the concrete language value:

```neva
stream<T> union {
    Open
    Data T
    Close
}
```

It is represented at runtime as `messages.UnionMsg`, but canonical construction and decoding are currently private helpers in `internal/runtime/funcs/stream_helpers.go`.

## Problem

Several runtime functions receive or construct scalar values and always materialize boxed containers:

| Path | Current output |
| --- | --- |
| `stream_to_list` | `NewListMsg([]Msg)` |
| `array_port_to_list` | `NewListMsg([]Msg)` |
| `stream_to_dict` | `NewDictMsg(map[string]Msg)` |
| `stream_zip_many` | `NewListMsg([]Msg)` for each zipped item |
| `os_environ` | boxes each string, then `NewListMsg([]Msg)` |

Thus a well-typed `stream<int>`, `[port] string`, or `stream<DictEntry<float>>` loses typed scalar storage at materialization. This defeats the typed-container fast paths introduced by #1127.

## Ownership decision

`messages` owns the **stream event value protocol**, because it only constructs, classifies, or decodes immutable `UnionMsg` values and imports no runtime transport code.

`runtime/funcs` owns **stream transport and lifecycle**: port receive/send, causes, concurrent coordination, waiting for `Open`, draining to `Close`, cancellation, and state machines such as Zip/Product. Do not move those state machines into `messages`.

Do not create a nested `messages/streams` package. Start with `internal/runtime/messages/stream.go`; a new package needs a real import/ownership boundary.

## Target API

Exact names can be confirmed against existing style, but the target shape is:

```go
func NewStreamOpenMsg() UnionMsg
func NewStreamDataMsg(data Msg) UnionMsg
func NewStreamCloseMsg() UnionMsg
func IsStreamOpen(msg Msg) bool
func IsStreamData(msg Msg) bool
func IsStreamClose(msg Msg) bool
func StreamDataValue(msg Msg) Msg // panics for an invalid internal stream event

func ListFromMessages(values []Msg) Msg
func DictFromMessages(values map[string]Msg) Msg
```

`ListFromMessages` / `DictFromMessages` return typed bool/int/float/string storage only when every value has that same scalar representation. Mixed, nested, and empty inputs remain untyped: without a value, empty runtime materialization has no scalar representation evidence. These are pure materializers; they must not mutate input slices/maps.

## Migration scope

1. Move the current stream union helper logic and direct tests from `funcs` to `messages/stream.go`.
2. Implement and benchmark the two container materializers in `messages`.
3. Migrate `stream_to_list`, `stream_to_dict`, `array_port_to_list`, and `stream_zip_many`.
4. Change `os_environ` to build `[]string` and return `NewListStringMsg`.
5. Audit all remaining direct `NewListMsg` / `NewDictMsg` calls in runtime functions. Keep calls producing non-scalar payloads (for example `DirEntry`, Zip/Product structs) explicitly untyped and document why in the PR.
6. Update developer runtime documentation with the stream value-protocol vs transport boundary.

## Tests and benchmarks

- Direct messages tests for every scalar kind, mixed/nested/empty fallback, and source-storage isolation.
- Direct benchmarks for typed scalar materialization and fallback.
- Runtime-function tests for unchanged stream lifecycle, public error text, and causes.
- At least one e2e case for stream-to-list and stream-to-dict scalar materialization.

## Non-goals

- No change to the public `stream<T>` union contract.
- No redesign of stream cancellation, stream errors, or indexing; those belong to #908 and related stream API issues.
- No speculative collection API beyond representation-aware materialization.

## Related

- #1148
- #1127
- #1155
- #908

emil14:
Closes #1156
Closes #908

## Summary

- move the `stream<T>` union value protocol into `internal/runtime/messages`; transport and lifecycle remain in runtime functions
- make `NewListMsg` and `NewDictMsg` representation-aware: homogeneous bool/int/float/string messages retain typed storage, while mixed, nested, and empty collections remain boxed
- expose explicit `NewUntypedListMsg` and `NewUntypedDictMsg` only for callers that deliberately require boxed storage
- migrate stream and array materializers to the representation-aware constructors; remove the former `funcs` stream helpers
- keep Neva JSON formatting independent of typed vs untyped dictionary storage
- update stream documentation and runtime-function ownership guidance

## Audit

The only remaining direct generic container constructor in runtime functions is `os_fs`, which emits `DirEntry` structs and therefore has no scalar typed representation. `list_to_stream` and `dict_to_stream` intentionally box at the container-to-stream boundary.

## Validation

- `go test ./...`
- targeted `golangci-lint` for `internal/runtime/messages` and `internal/runtime/funcs`
- materialization benchmarks

The local pre-push full lint is blocked by pre-existing findings in `internal/compiler/**`; this diff is clean under targeted lint.

Related: #1148, #1127, #1155


internal/runtime/messages/list.go
```diff
@@ -133,6 +133,90 @@ func NewListMsg(v []Msg) Msg {
 	return untypedListMsg{v: v}
 }
 
+// ListFromMessages materializes values into the most specific scalar list
+// representation available. Mixed, nested, and empty values remain untyped.
+//
+//nolint:ireturn // Msg contract type.
+func ListFromMessages(values []Msg) Msg {
```
emil14: 1. why not return ListMsg? it implements Msg
2. `NewListMsg` should be called `NewUntypedListMsg`, same naming pattern should apply for dict related code
3. Make sure "new..." constructors have doc-coments explaning that you should avoid using untyped containers whenever possible and what you should use instead in case X and Y

internal/runtime/messages/dict.go
```diff
@@ -133,6 +123,90 @@ func NewDictMsg(d map[string]Msg) Msg {
 	return untypedDictMsg{v: d}
 }
 
+// DictFromMessages materializes values into the most specific scalar dict
+// representation available. Mixed, nested, and empty values remain untyped.
+//
+//nolint:ireturn // Msg contract type.
+func DictFromMessages(values map[string]Msg) Msg {
```
emil14: maybe `NewListMsg` functon should do what this function is doing? and the same logic for list?

internal/runtime/messages/list.go
emil14: Да, согласен. Сделал API единообразным: `NewListMsg` теперь возвращает `ListMsg`, `NewDictMsg` — `DictMsg`; оба интерфейса уже реализуют `Msg`. Принудительно boxed-конструкторы переименованы в `NewUntypedListMsg` / `NewUntypedDictMsg`, а doc-comments объясняют, когда использовать обычный, scalar или explicit-untyped вариант.

internal/runtime/messages/dict.go
emil14: Да. Убрал отдельные `ListFromMessages` / `DictFromMessages`: их логику теперь выполняют `NewListMsg` / `NewDictMsg`. Поэтому compiler literals и все новые call sites автоматически сохраняют typed scalar storage, а explicit `NewUntyped…Msg` оставлен только для осознанного boxed-представления.

Issue #1171, opened 2026-08-01
RFC: add nevafmt as the canonical neva fmt formatter


emil14:
## Decision summary

Introduce **nevafmt** as Neva's canonical, first-party source formatter.

- The canonical command is **neva fmt**, shipped in every Neva release.
- The formatter lives in the main nevalang/neva repository and is versioned with the parser and language version.
- A separate neva-fmt executable may later be released for editor/package-manager integration, but it must be produced by the same release. neva tool fmt is not the primary workflow and is not required for v1.
- The formatter is deliberately narrow: parse source, preserve comments and meaning, and render canonical layout. It does not reorganize a program or perform semantic rewrites.
- neva lint remains a distinct tool and policy layer; see #1140.

This adopts Go as the default UX and scope reference. Departures are explicit and only where they solve a concrete workflow gap.

## Motivation

A formatter is part of the language experience, not merely an editor preference:

1. every developer, CI job, and editor gets the same result immediately after installing Neva;
2. formatter and parser versions cannot drift;
3. code review is not spent on whitespace;
4. canonical source supports examples, generated code, visual tooling, and future bootstrap work;
5. Neva has a graphical syntax surface with nested connections, so stable rendering is an important readability contract.

The existing neva tool <name> command proxies an independently installed neva-<name> binary. That is appropriate for LSP and optional integrations, but it creates an additional installation and a compatibility boundary. Canonical formatting must not depend on it.

## References and conclusions

| Reference | Relevant lesson | Neva decision |
| --- | --- | --- |
| [gofmt / go fmt](https://go.dev/doc/cmd) | One canonical formatter, narrow scope, direct file/stdin workflow, stable toolchain ownership. | Primary CLI and product reference. |
| [Deno fmt](https://docs.deno.com/runtime/reference/cli/fmt/) | --check is a clear CI contract. | Add -check; this is the only v1 UX departure from gofmt's core flags. |
| [zig fmt](https://ziglang.org/download/0.4.0/release-notes.html) | Canonical rendering is a language-toolchain responsibility; formatting can be implemented in the language over its own syntax infrastructure. | Keep self-hosting as a later implementation change, not a CLI change. |
| [Rust rustfmt / cargo fmt](https://doc.rust-lang.org/cargo/commands/cargo-fmt.html) | Formatting must be stable enough for CI; toolchain components may have version coupling. | Ship with Neva rather than require a separately installed component. |
[...]

PR #1172, opened 2026-08-01
Add nevafmt parser foundation

emil14:
## Summary

Starts the implementation of the canonical `neva fmt` RFC (#1171).

- adds a syntax-preserving `parser.ParseSource` API with original source, the complete ANTLR token stream, and parse tree;
- makes the existing compiler AST parser reuse that parsing setup;
- documents why formatting ships as first-party `neva fmt`, rather than an independently installed tool;
- adds coverage for comment/newline token retention and syntax-error reporting.

## Design boundary

This PR deliberately does **not** expose a partial `neva fmt` command or a whitespace heuristic. The next slice will define golden fixtures and a pure renderer on this API, so formatting cannot silently damage valid source layout or comments.

## Validation

- `go test ./internal/compiler/...`
- `go test ./internal/cli`
- `git diff --check`

The repository-wide pre-push lint currently reports 25 pre-existing findings in `internal/compiler/analyzer` and `internal/compiler/typesystem`, outside this PR's diff. Unit tests, gofix diff check, and vulncheck passed.

### DIFF
--- a/docs/user/qna.md
+++ b/docs/user/qna.md
@@ -64,30 +64,6 @@ In practice, this means:
 - Human-readable scalar-to-string formatting should also live in `strconv` (Go style),
   for example `strconv.Itoa`/`FormatFloat`/`FormatBool`.
 
-## Why will Neva ship `neva fmt` instead of requiring a separate formatter?
-
-Formatting source code is a first-party language-tooling contract, not a
-third-party convention. Neva will ship one canonical formatter as `neva fmt`
-with the main Neva release, so a developer, CI system, and editor use the same
-formatter version as the parser that accepts the source.
-
-This follows the Go model: formatting should work immediately after installing
-the language toolchain, should require no project-specific style configuration,
-and should remove whitespace-only review discussion. The formatter will parse
-one source file at a time; it will not require a successful build, resolved
-dependencies, or type checking.
-
-`neva tool` remains useful for separately installed developer tools such as a
-language server or experimental integrations. It is not the primary entry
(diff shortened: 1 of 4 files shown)


docs/user/qna.md
```diff
@@ -64,6 +64,30 @@ In practice, this means:
 - Human-readable scalar-to-string formatting should also live in `strconv` (Go style),
   for example `strconv.Itoa`/`FormatFloat`/`FormatBool`.
 
+## Why will Neva ship `neva fmt` instead of requiring a separate formatter?
```
emil14: давай в конец и покороче

internal/compiler/parser/syntax_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestParseSourcePreservesSourceTokensAndTree(t *testing.T) {
+	t.Parallel()
+
+	source := []byte("// Greets the caller\n" +
```
emil14: it actually does nothing

docs/user/qna.md
emil14: Done in 42c35592: moved the Q&A entry to the end of the file and reduced it to one short paragraph.

internal/compiler/parser/syntax_test.go
emil14: Done in 42c35592: renamed the fixture comment to `Passes start through.` so it matches the component's actual behavior.

PR #1174, opened 2026-08-01
Add nevafmt renderer

emil14:
## Summary

Continues RFC #1171 with the first pure renderer for syntax-valid Neva source.

- adds an internal formatter that depends only on `parser.ParseSource`;
- normalizes horizontal spacing, tab indentation, trailing whitespace, and CRLF to LF;
- preserves comments and existing logical line breaks;
- keeps import policy, declaration ordering, line-width wrapping, lint fixes, and CLI integration out of scope;
- adds golden fixtures for a component, a multiline struct literal, and dependency injection, plus idempotence, parser-error, and CRLF coverage.

## Design

The renderer annotates only syntax contexts that have distinct spacing rules: struct-value colons and node dependency-injection braces. It does not print from the semantic AST and does not invoke module resolution, analysis, desugaring, or code generation.

## Validation

- `go test ./internal/compiler/... ./internal/formatter ./internal/cli`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest run ./internal/formatter`
- `git diff --check`

The local repository-wide pre-push lint remains blocked by existing findings outside this PR; the focused formatter lint is clean.


pkg/formatter/testdata/basic/golden.neva
```diff
@@ -0,0 +1,4 @@
+// Passes the start signal through.
```
emil14, line 1: im not sure if fmt must be a intenral level directory, how about moving it under internal/cli/fmt/...? wdyt?

pkg/formatter/testdata/dependency_injection/golden.neva
```diff
@@ -0,0 +1,5 @@
+def Main(start any) (stop any) {
```
emil14, line 1: all this feels at a glance like e2e tests

maybe better place would be under https://github.com/nevalang/neva/tree/main/e2e/cli ?

pkg/formatter/testdata/dependency_injection/input.neva
```diff
@@ -0,0 +1,5 @@
+def Main(start any)(stop any){
```
emil14, line 1: just so you know as an alternative to have all this as *.neva files it could be just string literals insidego tests

im not saying its better just fyi

internal/compiler/formatter/formatter_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestFormatGolden(t *testing.T) {
+	t.Parallel()
+
+	cases, err := os.ReadDir("testdata/format")
```
emil14: ok well maybe it's not that bad to have them as files

and this is not actually e2e

pkg/formatter/testdata/dependency_injection/input.neva
emil14: Agreed that string literals are viable. I kept paired `.neva` files because source/golden diffs stay readable and the corpus can grow by syntax production without turning the Go test into a large embedded document.

internal/compiler/formatter/formatter_test.go
emil14: Yes — retained as golden unit fixtures rather than e2e coverage.

pkg/formatter/testdata/dependency_injection/golden.neva
emil14: Kept these as unit-level golden fixtures. They exercise the pure source-to-source renderer with no CLI, filesystem traversal, or process boundary, so `e2e/cli` would conflate the renderer contract with future command behavior.

pkg/formatter/testdata/basic/golden.neva
emil14: Moved in 46ec7ddd to `internal/compiler/formatter`. It is syntax tooling coupled to the parser, not CLI behavior; a future `neva fmt` command should call it. If cross-repository LSP reuse needs an exported shared module, that extraction deserves its own proposal rather than placing the engine under CLI.

pkg/formatter/testdata/basic/input.neva
```diff
@@ -0,0 +1,4 @@
+// Passes the start signal through.
```
emil14, line 1: is there a reason why its testdata/format/basic instead of just testdata/basic?

pkg/formatter/formatter.go
```diff
@@ -0,0 +1,203 @@
+// Package formatter renders syntax-valid Neva source into its canonical layout.
+package formatter
```
emil14, line 2: what do you think about having this inside pkg/formatter instead of compiler/formatter? would we benefit from it? ASking because its not really part of the compilation pipeline and actually going to be part of the CLI

pkg/formatter/formatter.go
```diff
@@ -0,0 +1,203 @@
...
+		return nil, err
+	}
+
+	return FormatParsed(parsed), nil
+}
+
+// FormatParsed renders syntax-valid parsed source. It has no dependency on
+// module resolution, analysis, desugaring, or code generation.
+func FormatParsed(parsed parser.ParsedSource) []byte {
+	if len(parsed.Tokens) == 1 {
```
emil14, line 28: it would be nice to have short single line comments on top of code-blocks inside this function to understand algorithm quickly

pkg/formatter/formatter.go
```diff
@@ -0,0 +1,203 @@
...
+		output.WriteString(renderLine(tokens, layout))
+		output.WriteByte('\n')
+
+		depth += braceDelta(tokens)
+	}
+
+	return []byte(output.String())
+}
+
+type indexedToken struct {
```
emil14, line 66: absolutely EVERY ENTITY must have comments...

pkg/formatter/formatter.go
emil14: Done in e4b00383: added short comments for syntax annotation, token grouping, blank-line handling, per-line indentation, and brace-depth propagation.

pkg/formatter/testdata/basic/input.neva
emil14: Agreed — simplified in e4b00383 to `testdata/<case>`.

pkg/formatter/formatter.go
emil14: Done in e4b00383: documented every production type, helper function, and listener method in the formatter.

pkg/formatter/formatter.go
emil14: Agreed — moved to `pkg/formatter` in e4b00383. It better represents the reusable formatting API; the future CLI should be a thin caller. This does not yet make it importable by the separate LSP repository because the parser facade remains internal; that extraction should be a separate, deliberate change.

Issue #1160, opened 2026-07-28
Implement generic resolved-type binding for extern components


emil14:
## Goal

Add one generic low-level compiler capability: an extern component may request the compiler-resolved form of one type expression as its static runtime config.

Working syntax:

```neva
#extern(runtime_function_name)
#bind_type(TypeExpr)
pub def Component<T>(...)
```

This capability is generic. The compiler must not know about JSON or any standard-library package name.

## Contract

- `TypeExpr` is any valid type expression resolvable in component scope, not only a declared type parameter.
- `#bind_type` is valid only on a component declaration that also has `#extern`.
- It may appear at most once.
- Node-level use is invalid.
- V1 binds exactly one resolved type; multiple bindings are deferred until a real consumer requires them.
- It may not compete with another producer for the single existing `FuncCall.Config` message.
- Analyzer rejects invalid placement, duplicates, incompatible combinations, unknown names, and unresolved expressions.
- Lowering receives the fully resolved type and emits one ordinary message matching the portable `std/reflect.Type` wire shape.
- IR/generated Go contain only that message; they do not reference the Neva `reflect` package at runtime.
- Static config invariants are compiler-owned. Runtime-function `Create` may fail fast on a compiler defect.

The first consumers are both `encoding/json.Marshal<T>` and `Unmarshal<T>`. Marshal also needs the descriptor because runtime union values do not contain their complete static union shape.

## Dependency

#1163 must land first. The current directive grammar accepts only one identifier argument, while this contract requires a structured arbitrary type expression and forbids late string reparsing.

## Verification

- parser/AST coverage from #1163;
- analyzer positive/negative tables for every invariant;
- lowering tests for scalars, `any`, empty typed containers, nested composites, tagged unions, aliases, and recursion;
- IR/backend/e2e contract tests;
- developer documentation explaining directives as low-level compiler contracts.

Broader directive renaming and user-defined Go interop remain separate design work; see #976.

emil14:
## JSON-derived requirement

The likely mechanism is conceptually a type binding for an extern component: the compiler resolves a selected generic type parameter and emits its JSON-relevant resolved structural representation into the existing static config message. This is not `#bind`: `#bind` attaches a constant to a node, while the new capability is declared by a component and instantiated from its resolved type arguments.

The final directive name and syntax are intentionally unresolved. Do not make the compiler aware of the `json` package.

Analyzer invariants to design:

- only valid on an extern component;
- refers to declared type parameter(s);
[...]

Issue #933, opened 2025-10-26
Create `json` package design proposal


emil14:
Use json `v2` under the hood

Related to #28 and #790

emil14:
Related follow-ups created during the design discussion:

- #1159 separates streaming JSON decoding from ordinary `Marshal`/`Unmarshal`: framing, cancellation, and stream error semantics are deferred and must reuse the same per-value mapping.
- #1160 tracks the directive-model refactor: generic resolved-type metadata for decoding must be a compiler capability, not JSON-specific knowledge.

Neither issue changes the first JSON package scope by itself.

emil14:
## Design record: static decoding bridge

The first `json` package remains deliberately narrow:

```neva
pub def Marshal<T>(data T) (res bytes, err error)
pub def Unmarshal<T>(data bytes) (res T, err error)
```

`Unmarshal<T>` needs the compiler-resolved structural form of `T`; an immutable template value is not a valid schema because empty containers and tagged-union instances erase required information. The compiler should place that information in the existing static `FuncCall.Config` as an ordinary Neva composite message. This does **not** add a new `Msg` implementation.

The config describes the resolved JSON-relevant shape only: scalars, bytes, list item type, dict item type, struct fields, and tagged-union members. It intentionally omits constraints, type parameters, and source aliases.

### Contract

For JSON-representable `v : T`, after finalizing `null`, `any`, bytes, and non-finite-float policy:

```text
Unmarshal<T>(Marshal(v)) == v
```

`Marshal(Unmarshal<T>(input))` preserves JSON semantics but may normalize bytes. A document may validly decode into a different source alias if its resolved structural target shape accepts that document.

### Deliberately deferred

- `any` and `null` policy;
- public runtime type values and type assertion/introspection (#749);
- dynamic non-stream decoding (#1164);
- streaming decoding/framing (#1159).

### Important open representation detail

A direct recursive type-value tree handles non-recursive schemas, but cannot finitely represent a recursive source schema such as `error -> maybe<error>`. The design must decide between a self-contained descriptor graph with references, or an equivalent named-reference/definition mechanism. This is a representation decision, not a reason to attach type metadata to every runtime message.

### Validation expectation

Add compiler-lowering tests for resolved type to config message, runtime-function unit tests for descriptor-guided decoding, e2e round-trip cases, and targeted benchmarks. Runtime may rely on compiler-generated static config invariants and fail fast on compiler defects; user-provided dynamic descriptors are a separate future boundary.

[...]

Issue #749, opened 2024-11-04
Type-assertion


emil14:
Related to #747 #726 #725

emil14:
# FYI TypeScript

TS just like Nevalang doesn't have compile-time (complex) type information at runtime, so the only way to e.g. assume if something is of specific shape is by raw trying to access fields

```ts
interface Cat {
    name: string;
    lives: number;
}

interface Dog {
    name: string;
    barks: boolean;
}

// Example data from a third-party library
const animals: (Cat | Dog)[] = [
    { name: 'Whiskers', lives: 9 },  // Cat
    { name: 'Fido', barks: true },    // Dog
];

// Type guard for Cat
function isCat(animal: any): animal is Cat {
    return (animal as Cat).lives !== undefined;
}

// Type guard for Dog
function isDog(animal: any): animal is Dog {
    return (animal as Dog).barks !== undefined;
}

// Function to handle animals
function handle(animal: Cat | Dog) {
    if (isCat(animal)) {
        console.log(`This is a cat named ${animal.name} with ${animal.lives} lives.`);
    } else if (isDog(animal)) {
        console.log(`This is a dog named ${animal.name} that barks: ${animal.barks}.`);
    } else {
        console.error("Unknown animal type");
    }
}

// Example usage
for (const animal of animals) {
    handle(animal);
}
```

We either have to follow same approach or add `typesystem` information to runtime

emil14:
Some dump from editor

---

Solving problem of asserting a specific type on message of type `any`

```
def OneOf(data any) (x int, y models.User, z any) {
    data -> type -> switch {
        (type int) -> x
        (type models.User) -> y
        _ -> z
    }
}
```

## Parser/Sourcecode/Analyzer

- New `type` sender. It's a special kind of sender that receives the message and sends its type-id. It's expected to be used like a single-port per side node and analyzer must check that.
- Another one `type_v2` sender. It's a completely different syntax and implementation. It's in `(type ...)` like binary/ternary operation expression (but it's not) and its right part is an entity reference, that must refer to a type-definition
- Extend `core.Meta` with optional field `extra any` and put type-id there at parsing step. Create global counter and increment it for each type definition in the program.

### Desugaring/Irgen

### Runtime

1. make each `runtime.NewXXXMsg(...)` constructor take a type-id as an argument. Example: `runtime.NewIntMsg(42, 238475)`.
2. add `typeID uint64` field to `runitme.internalMsg`

???

emil14:
[...]

PR #1175, opened 2026-08-02
Add portable std/reflect type descriptor

emil14:
## Summary

- add public `std/reflect.Type` and `TypeNode` as a finite flat indexed graph
- add canonical message construction/reading with validation for empty, malformed, duplicate, and out-of-bounds descriptors
- cover recursive back-edges with Go round-trip tests and a Neva e2e contract
- fix two existing compiler blockers exposed by the descriptor: finite nested base containers were misclassified as direct recursion, and union constant payloads were lowered with the whole union type
- commit the compaction-safe JSON implementation plan and document the narrow reflection substrate

This PR deliberately does not add `TypeOf`, `#bind_type`, JSON logic, per-message metadata, or a global type registry. It is iteration 2 of #933 and the prerequisite representation for #1160; it also records the narrow substrate relevant to #749.

## Wire contract

- `reflect.Type = list<TypeNode>`
- node `0` is the root
- every composite edge is an integer index into the same list
- `StructField = { name string, node int }`
- `UnionCase = { tag string, data maybe<int> }`
- aliases, constraints, and generic parameters are absent

## Validation

- `go test ./internal/compiler/... ./internal/runtime/messages ./e2e/reflect_type_descriptor`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest run --new-from-rev=origin/main`
- `git diff --check`

The repository-wide local pre-push lint still reports pre-existing strict-lint findings in compiler packages; the changed-lines lint reports 0 issues.


.codex/plans/encoding_json_v1.md
```diff
@@ -0,0 +1,393 @@
+# `std/encoding/json` v1 implementation plan
```
emil14: lets not merged this here just keep in local and in memory

docs/developer/runtime-functions.md
```diff
@@ -98,6 +98,24 @@ have the same scalar representation; mixed, nested, and empty collections
 remain untyped. Use `NewUntypedListMsg` or `NewUntypedDictMsg` only when boxed
 storage is deliberately required.
 
+## Resolved Type Descriptors
```
emil14: Sort sections by their importancy, the most common stuff first and more specific is later

e2e/reflect_type_descriptor/main/main.neva
```diff
@@ -0,0 +1,19 @@
+import { reflect }
```
emil14: didn't we close issue about forbidding single line imports? all imports syntaticaly must always be 

import {
...
}

if not, do it now

std/reflect/reflect.neva
```diff
@@ -0,0 +1,29 @@
...
+
+// TypeNode describes one node in a resolved type graph.
+pub type TypeNode union {
+	Any
+	Bool
+	Int
+	Float
+	String
+	Bytes
+	List int
```
emil14, line 17: explain how indexing works in this file

int in List, Dict and inside SturctField and UnionCase

.codex/plans/encoding_json_v1.md
emil14: Removed the plan from this PR. The identical local copy remains at `.codex/plans/encoding_json_v1.md` in the main checkout, and I recorded that it must stay local rather than being committed.

docs/developer/runtime-functions.md
emil14: Moved `Resolved Type Descriptors` after the general execution, container, stream, concurrency, and state sections, immediately before tests/comments. The document now progresses from common runtime-function concerns to this narrower compiler/runtime bridge.

e2e/reflect_type_descriptor/main/main.neva
emil14: Confirmed: PR #1177 already made multiline import blocks mandatory in the grammar and parser tests. Rebasing this branch onto current `main` brought that rule in, and this fixture now uses the required multiline form.

std/reflect/reflect.neva
emil14: Expanded the public comments: every integer is a zero-based index into the same `Type` list; node 0 is the root; `List`/`Dict` point to their item node, `StructField.node` points to the field type, and `UnionCase.data` is `None` for tag-only or `Some(index)` for a payload. I also added the concrete `list<string> = [TypeNode::List(1), TypeNode::String]` example and noted forward, backward/recursive, and shared edges.

e2e/reflect_type_descriptor/main/main.neva
```diff
@@ -0,0 +1,21 @@
...
+	reflect.TypeNode::Union([
+		{ tag: 'Some', data: maybe<int>::Some(0) },
+		{ tag: 'None', data: maybe<int>::None }
+	])
+]
+
+def Main(start any) (stop any) {
+	pass Pass<reflect.Type>
+	---
+	:start -> $recursive -> pass -> :stop
```
emil14: what exactly does this test is testing...? how about making it useful?

docs/developer/runtime-functions.md
```diff
@@ -120,6 +120,24 @@ permits many instances.
 Prefer an explicit public runtime API or a dependency passed through an
 existing runtime boundary over package-level mutable singletons.
 
+## Resolved Type Descriptors
```
emil14, line 134: This is the least important section, it's very specific

internal/compiler/irgen/message.go
```diff
@@ -56,7 +65,17 @@ func getIRMsgBySrcRef(
...
+				return nil, errors.New("union message requires resolved union type")
+			}
+			payloadType, found := typeExpr.Lit.Union[constant.Message.Union.Tag]
+			if !found || payloadType == nil {
+				return nil, fmt.Errorf(
+					"union payload type not found for tag %q",
+					constant.Message.Union.Tag,
+				)
+			}
+			dataMsg, err := getIRMsgBySrcRef(*constant.Message.Union.Data, scope, *payloadType)
```
emil14, line 78: there's something I do not understand

how exactly do we pass the type to IR? I dont see it... 

In my view it should be a type expr turning into a IR message that is assigned to func config msg but where?

internal/runtime/messages/reflection.go
```diff
@@ -0,0 +1,354 @@
...
+	Data *int64
+	Tag  string
+}
+
+// ReflectTypeToMessage builds the ordinary message representation of a
+// std/reflect.Type value.
+//
+//nolint:ireturn // Msg is the value-layer contract.
+func ReflectTypeToMessage(value ReflectType) (Msg, error) {
+	if err := reflectTypeValidate(value); err != nil {
```
emil14: are you sure we should validate this? if this is check of compiler its not needed

internal/runtime/messages/reflection.go
```diff
@@ -0,0 +1,354 @@
...
+type ReflectUnionCase struct {
+	Data *int64
+	Tag  string
+}
+
+// ReflectTypeToMessage builds the ordinary message representation of a
+// std/reflect.Type value.
+//
+//nolint:ireturn // Msg is the value-layer contract.
+func ReflectTypeToMessage(value ReflectType) (Msg, error) {
```
emil14: If we know for sure then lets return concrete messages like ListMsg

internal/runtime/messages/reflection.go
```diff
@@ -0,0 +1,354 @@
...
+	nodes := make([]Msg, len(value.Nodes))
+	for i := range value.Nodes {
+		nodes[i] = reflectTypeNodeToMessage(value.Nodes[i])
+	}
+	return NewUntypedListMsg(nodes), nil
+}
+
+// ReflectTypeFromMessage reads and validates an ordinary std/reflect.Type
+// message without retaining its backing message storage.
+func ReflectTypeFromMessage(value Msg) (ReflectType, error) {
```
emil14: lets just accept ListMsg and do not care about other Msgs here

internal/runtime/messages/reflection.go
```diff
@@ -0,0 +1,354 @@
...
+				NewStructField("data", data),
+			})
+		}
+		return NewUnionMsg(string(node.Kind), NewUntypedListMsg(cases))
+	default:
+		panic("unexpected validated reflect type kind")
+	}
+}
+
+func reflectTypeNodeFromMessage(value Msg) (ReflectTypeNode, error) {
```
emil14, line 122: every fucking enitty sycbh as type, function or method must have doc comments

e2e/reflect_type_descriptor/main/main.neva
emil14: Made the e2e observable. It now sends the descriptor through `Pass<reflect.Type>`, prints it, and asserts the exact JSON wire value: Struct field indexes, Union cases, `Some(0)` recursive back-edge, and tag-only `None` must all survive unchanged.

docs/developer/runtime-functions.md
emil14: Moved `Resolved Type Descriptors` to the end of runtime-functions.md, after the testing/comment guidance. It is now the final, most specific section.

internal/compiler/irgen/message.go
emil14: You do not see that bridge because this PR deliberately does not implement it yet. This PR only establishes the portable public `reflect.Type` wire contract and its message conversion. The existing `getConfigMsg` lowers only node `#bind` constants into `ir.FuncCall.Msg`. PR #1160 will add `#bind_type`: analyzer resolves its TypeExpr, IR generation builds the corresponding `reflect.Type`-shaped IR message, and assigns it to that same `FuncCall.Msg` field. I added this distinction directly above `getConfigMsg`.

internal/runtime/messages/reflection.go
emil14: Removed validation from `ReflectTypeToMessage`. Its native input is compiler-produced, so an invalid descriptor is an internal compiler invariant violation rather than a recoverable runtime condition. `ReflectTypeFromMessage` still validates the already-typed wire list while decoding it into the native graph.

internal/runtime/messages/reflection.go
emil14: Changed `ReflectTypeToMessage` to return `ListMsg`, the precise list-level message contract, instead of the generic `Msg`.

internal/runtime/messages/reflection.go
emil14: Changed `ReflectTypeFromMessage` to accept `ListMsg` directly. The irrelevant top-level `Msg` assertion and its test are gone.

internal/runtime/messages/reflection.go
emil14: Added doc comments to every function in reflection.go, including the private encoders, decoders, and validators; exported types and functions already have symbol comments.

PR #1176, opened 2026-08-02
docs: add papercut workflow

emil14:
## Summary
- add a developer journal for small repository and harness friction
- add the repository-local `papercut` capture/triage skill
- record two observed harness frictions as its initial entries

## Validation
- `git diff --check`
- `lefthook run pre-commit`
- pre-push `make test-unit` and `make vulncheck` passed
- pre-push `make lint` failed on violations unchanged from `origin/main`, including reports from a sibling worktree; GitHub CI is the readiness gate

The existing untracked `.codex/plans/encoding_json_v1.md` was not included.

### DIFF
--- a/.codex/skills/papercut/SKILL.md
+++ b/.codex/skills/papercut/SKILL.md
@@ -1,0 +1,41 @@
+---
+name: "papercut"
+description: "Capture or triage small Neva repository and engineering-harness frictions without derailing feature work."
+---
+
+# Papercut
+
+Use this skill when a person or agent encounters minor friction while doing
+other Neva work: a flaky command, stale cache, misleading error, obsolete
+script, or documentation that contradicts observed behavior.
+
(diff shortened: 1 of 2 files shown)


.codex/skills/papercut/SKILL.md
```diff
@@ -0,0 +1,41 @@
...
+   What I was doing -> what got in the way. Include the relevant command or
+   error, workaround, and likely cause or fix when known.
+   ```
+
+3. Record only evidence already observed. Never put credentials, access tokens,
+   or private data in an entry.
+
+## Triage
+
+1. Read the full journal and group duplicate symptoms.
```
emil14, line 33: Actually its better to have `.codex/papercut/<YYYY-MM-DD>.md` with yaml frontmatter with field `description: ...` as a summary of the problem described, so that LLM does not have to read the full papercut all the time and instead can just read summaries and dive-in if needed

PR #1182, opened 2026-08-02
Add bind type compiler bridge

emil14:
Implements #1160.

- validates component-only `#bind_type(TypeExpr)` on `#extern` declarations and rejects conflicts with node `#bind`;
- resolves the expression against call-site generic arguments and emits the established `std/reflect.Type` wire shape through `ir.FuncCall.Msg`;
- adds deterministic structural lowering, nested-container coverage, a recursive back-edge builder test, and IR/e2e contract coverage;
- documents the compiler boundary.

Validation:
- `go test ./internal/compiler/analyzer ./internal/compiler/irgen ./e2e/bind_type`
- `golangci-lint run --new-from-rev=origin/main ./internal/compiler/analyzer ./internal/compiler/irgen ./e2e/bind_type`


e2e/bind_type/main/main.neva
```diff
@@ -0,0 +1,10 @@
+#extern(int_inc)
```
emil14, line 1: um... so `#extern` and `#bind_type` are available outside of the std module? is this because (extern I mean) of the Go-interop?

internal/compiler/irgen/irgen.go
```diff
@@ -4,12 +4,15 @@ import (
...
 	"github.com/nevalang/neva/internal/compiler/ir"
+	ts "github.com/nevalang/neva/internal/compiler/typesystem"
 	"github.com/nevalang/neva/pkg"
 	src "github.com/nevalang/neva/pkg/ast"
 	"github.com/nevalang/neva/pkg/core"
 )
 
-type Generator struct{}
+type Generator struct {
+	resolver ts.Resolver
```
emil14, line 14: why do we need resolver here? irgen assumes program it consumes is analyzed by analyzer and therefore all the types inside are resolved already

internal/compiler/analyzer/component.go
```diff
@@ -77,9 +81,53 @@ func (a Analyzer) analyzeComponent(
...
+//nolint:gocritic // Analyzer pipeline values are passed by value consistently.
+func (a Analyzer) validateComponentDirectives(
+	component src.Component,
+	iface src.Interface,
+	scope src.Scope,
+	hasExtern bool,
+) *compiler.Error {
+	if component.Directives.Has(src.DirectiveKindBind) {
+		return &compiler.Error{
+			Message: "#bind directive is only valid on a component node",
```
emil14, line 105: #bind directive is only valid on a component node, not the component itself

internal/compiler/analyzer/component.go
```diff
@@ -77,9 +81,53 @@ func (a Analyzer) analyzeComponent(
...
+		}
+	}
+
+	bindType, hasBindType := component.Directives.Find(src.DirectiveKindBindType)
+	if !hasBindType {
+		return nil
+	}
+	if !hasExtern {
+		return &compiler.Error{
+			Message: "#bind_type directive requires #extern",
```
emil14, line 116: 1. `#bind_type directive requires #extern because the type should be passed to ruintime function that implements the native component`
2. create GH issue about compiler errors in general, or find an existing one, the idea is that we must do a full revision and figure our (and document in repo dev docs) compiler errors policy - tone, style, etc. The main idea is that compiler errors should follow languages like Rust and Elm, where error is easy to grasp. This not only includes "what happened" and "where the problem is" but also providing a little extra information so that the user is unsdersanding hte language better (and LLM too) and also (but only where clear) a hint of what can be done instead (one or more options)

internal/compiler/analyzer/nodes.go
```diff
@@ -331,6 +337,12 @@ func (a Analyzer) getInterfaceAndOverloadingIndexForNode(
 			Meta:    entity.Meta(),
 		}
 	}
+	if hasBind && version.Directives.Has(src.DirectiveKindBindType) {
+		return src.Interface{}, nil, &compiler.Error{
+			Message: "#bind cannot be used with a component that declares #bind_type",
```
emil14, line 342: shortly explain why

internal/compiler/irgen/func.go
```diff
@@ -22,28 +25,81 @@ func (Generator) getFuncRef(versions []src.Component, node src.Node) (string, sr
...
+	component src.Component,
+	callerScope src.Scope,
+	componentScope src.Scope,
+) (*ir.Message, error) {
 	bind, hasBind := node.Directives.Find(src.DirectiveKindBind)
-	if !hasBind {
-		//nolint:nilnil // nil config is expected when no bind directive is present
+	bindType, hasBindType := component.Directives.Find(src.DirectiveKindBindType)
+
+	switch {
```
emil14, line 44: what if it has both?

internal/compiler/irgen/func.go
```diff
@@ -22,28 +25,81 @@ func (Generator) getFuncRef(versions []src.Component, node src.Node) (string, sr
...
+			return nil, err
+		}
+
+		return getIRMsgBySrcRef(
+			entity.Const.Value,
+			callerScope.Relocate(location),
+			entity.Const.TypeExpr,
+		)
+	case hasBindType:
+		resolvedType, err := g.resolveBoundType(
```
emil14, line 58: it must already be resolved by now

PR #1186, opened 2026-08-04
Add neva fmt command

emil14:
## Summary

- add the first-party `neva fmt` command over the syntax-only formatter
- support stdin, files, deterministic recursive directories, `-w`, `-d`, `-l`, `-check`, and `-e`
- write `-w` changes atomically while preserving permissions
- define CLI exit codes and document the command, boundaries, and directory skip-list

## Validation

- `go test ./internal/cli ./cmd/neva ./pkg/formatter ./internal/compiler/parser/...` — 108 passed
- `golangci-lint run ./cmd/neva ./internal/cli ./pkg/formatter ./internal/compiler/parser/...` — 0 issues
- built binary: `fmt -check` exits 1 and conflicting output modes exit 2

The local repository-wide pre-push lint traverses an unrelated sibling worktree and reports its existing issues. GitHub CI is authoritative.


cmd/neva/main.go
```diff
@@ -1,9 +1,12 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"os"
 
+	urfavecli "github.com/urfave/cli/v2"
```
emil14: I think we would want to hide urfave clu under the internal/cli pkg and leak here if possible

cmd/neva/main.go
```diff
@@ -35,9 +38,28 @@ func main() {
 
 	// run CLI app
 	if err := app.Run(os.Args); err != nil {
-		if _, err := fmt.Fprintln(os.Stderr, err); err != nil {
-			panic(err)
-		}
-		os.Exit(1)
+		exitWithError(err)
```
emil14, line 38: explain this change

docs/user/formatter.md
```diff
@@ -0,0 +1,65 @@
+# Formatter
```
emil14: this should be part of the tutorial, maybe CLI sesion or something, I need uyou to reason about this deeply and create perfect structure based on refs from Go, Rust, Zig, Kotlin, Odin, and V

Also, we ONLY need this section (as of now) if we sure its not going to change. E.g. if we looked at refs formatters and we are sure its how it works there kinda, and its not going to probably change

internal/cli/fmt_test.go
```diff
@@ -0,0 +1,170 @@
+package cli
```
emil14, line 1: is this real shell/fs level test? asking becayse we used to have e2e/cli for this kind of stuff but now Im questioning myself if we need too. wdyt?

internal/cli/fmt.go
```diff
@@ -0,0 +1,332 @@
...
+
+	"github.com/pmezard/go-difflib/difflib"
+	cli "github.com/urfave/cli/v2"
+
+	"github.com/nevalang/neva/pkg/formatter"
+)
+
+const nevaSourceExtension = ".neva"
+
+type fmtMode uint8
```
emil14: use custom type

internal/cli/fmt.go
```diff
@@ -0,0 +1,332 @@
...
+}
+
+func closeAndRemoveTemporaryFile(file *os.File, path string, cause error) error {
+	if err := file.Close(); err != nil {
+		cause = fmt.Errorf("%w; close temporary file %s: %w", cause, path, err)
+	}
+	return removeTemporaryFile(path, cause)
+}
+
+func removeTemporaryFile(path string, cause error) error {
```
emil14, line 378: do not forget to add doc comments to every func

cmd/neva/main.go
emil14: Done in 4cb02120. `cmd/neva` now depends only on `internal/cli`; the command framework and exit-status handling are fully contained there.

cmd/neva/main.go
emil14: This enables `neva fmt -check` to return status 1 for formatting drift and status 2 for an invalid invocation. The mapping now lives in `internal/cli` (`ExitCode`), so `cmd/neva` only prints the error and exits with the CLI-owned status.

docs/user/formatter.md
emil14: Reworked in 4cb02120. I removed the standalone page and placed the stable user workflow in the Tutorial's Welcome/CLI flow: stdin for editors, `-w`, `-d`, and `-check`. The shape follows the durable common ground: gofmt's stdin/file/directory model and preview/write modes ([Go](https://go.dev/cmd/gofmt/)); `cargo fmt --check` for CI ([Rust](https://doc.rust-lang.org/cargo/commands/cargo-fmt.html)); canonical compiler-owned formatting in Zig ([Zig](https://ziglang.org/download/0.4.0/release-notes.html)); and `v fmt -w` as the save/commit workflow ([V](https://docs.vlang.io/tools.html)). Kotlin is deliberately not the CLI model: its official formatter is IDE/style-guide oriented and configurable ([Kotlin](https://kotlinlang.org/docs/coding-conventions.html)). Detailed evolving mechanics stay in `neva fmt --help` and RFC #1171 rather than a duplicated reference page.

internal/cli/fmt_test.go
emil14: The previous tests use a real filesystem (`t.TempDir` and actual reads/writes) but call the command implementation in-process, so they are unit tests of traversal, modes, and atomic writes. I kept that fast layer and added `e2e/cli/fmt`: it builds and executes the real `neva` binary, verifies stdin, `-w`, `-check` exit 1/0, and invalid flags exit 2.

internal/cli/fmt.go
emil14: Done in 4cb02120: `fmtMode` is now a named string type with explicit mode values; its zero value remains the normal stdout mode.

internal/cli/fmt.go
emil14: Done in 4cb02120. Every production formatter helper and the new CLI exit-status helper now has a doc comment.

internal/cli/fmt.go
```diff
@@ -0,0 +1,352 @@
...
+	formatted []byte
+}
+
+// newFmtCmd creates the first-party source formatting command.
+func newFmtCmd() *cli.Command {
+	return &cli.Command{
+		Name:      "fmt",
+		Usage:     "Format Neva source files",
+		ArgsUsage: "[file or directory ...]",
+		Description: "With no paths, formats one source file from standard input. " +
```
emil14, line 52: - what is .neva?
- node_modules is an interesting catch, but should we hardcode such a specific path in our CLI? Interesting how other formatters are wortking with this?

internal/cli/fmt.go
```diff
@@ -0,0 +1,352 @@
...
+	default:
+		return false
+	}
+}
+
+// formatFiles formats paths concurrently and retains their input order.
+func formatFiles(paths []string) []fmtResult {
+	results := make([]fmtResult, len(paths))
+	jobs := make(chan int)
+	workers := min(runtime.GOMAXPROCS(0), len(paths))
```
emil14, line 273: aren't we limiting concurrency to one core by calling runtime.GOMAXPROCS(0)?

why do we need both wg wait and jobs? why not use wg.Go(...) for simpler syntax?

internal/cli/fmt.go
```diff
@@ -0,0 +1,352 @@
...
+	for index := range paths {
+		jobs <- index
+	}
+	close(jobs)
+	group.Wait()
+	return results
+}
+
+// formatFile reads and formats one source file without writing it.
+func formatFile(path string) fmtResult {
```
emil14, line 290: Why not orchestrate code concurrently like this?

```pseudo
for every file in files do
  go() { read -> format -> write }()
wait for all files
```

is there a meaming to imlement "transaction"? (Meaning "format all in memory before attempting to write")

Also if so, is real transaction implemented (write all or nothing)? If needed

internal/cli/fmt.go
```diff
@@ -0,0 +1,352 @@
...
+	return nil
+}
+
+// applyFmtResults emits or writes completed formatting results in input order.
+func applyFmtResults(paths []string, results []fmtResult, options fmtOptions, output io.Writer) (bool, error) {
+	changed := false
+	var reported []error
+	for index := range results {
+		result := &results[index]
+		resultChanged, resultErr := applyFmtResult(paths[index], result, options.mode, output)
```
emil14, line 167: why don't we apply all results concurrentyly?

docs/user/tutorial.md
```diff
@@ -150,6 +151,38 @@ Once again you should see `Hello, World!`.
 
 > Execute `neva build --help` to learn more - how to compile to Go, WASM or how to do cross-compilation e.g. compile linux binaries in windows.
 
+### Formatting Source
```
emil14: I think there's no need to dive deep into neva fmt cmd here, at best we could have a CLI common section where we tell about (if we didnt yet) help and a few most useful commands

emil14: Adjusted in eabbc4bc. The tutorial now has only a compact Command-line Help section for neva --help, command --help, neva run, and neva build; it no longer documents neva fmt mechanics.

internal/cli/fmt.go
emil14: Fixed in eabbc4bc. Those were speculative ecosystem-specific exclusions; .neva was not documented as a formatter boundary either. A directory argument now includes every .neva file under it and skips only Git metadata (.git).

internal/cli/fmt.go
emil14: Formatting is concurrent. Applying results remains ordered because several modes write one shared stdout stream, where concurrent writes would interleave and make output/error order nondeterministic. For -w, each file replacement is atomic; parallelizing those small rename operations does not justify losing deterministic failure order.

internal/cli/fmt.go
emil14: GOMAXPROCS(0) returns the current maximum number of OS threads executing Go code simultaneously; its default is the machine's available CPUs, not one core. I kept that bounded worker count to cap file descriptors and memory for large trees, and changed the worker lifecycle to WaitGroup.Go in eabbc4bc.

internal/cli/fmt.go
emil14: There is intentionally no cross-file transaction: each -w replacement is atomic for that file, but a filesystem cannot make multiple renames all-or-nothing. The bounded worker pool prevents unbounded file descriptors and memory; results are applied in input order so stdout, diffs, lists, and reported errors stay deterministic. This does mean -w is per-file atomic, not all-files atomic, and the command does not claim otherwise.

internal/cli/fmt.go
emil14: GOMAXPROCS(0) not only returns previous value but also sets the new one AFAIK check pelase

internal/cli/fmt.go
emil14: So should we introduce full concurrency for the modes where this is possible?

internal/cli/fmt.go
emil14: > The bounded worker pool prevents unbounded file descriptors and memory

Ok fair enough but make sure its reflected in the comments

> results are applied in input order so stdout, diffs, lists, and reported errors stay deterministic

Im trying to understand why cant we process each file concurrently, not just format but apply too. What exactlyu problem we going to get then.

emil14: Implemented in 752b0181. The command now formats all files through the bounded pool, reports every read/parse error before any -w operation, then uses the same bounded pool to atomically replace changed files. stdout modes remain ordered because they share one stream. A filesystem error during the write phase can still leave earlier files written, so this is deliberately a validation barrier plus per-file atomicity, not a false cross-file transaction.

internal/cli/fmt.go
emil14: Checked against the current Go documentation: GOMAXPROCS sets a new limit only when n is at least 1; with n < 1 it does not change the setting and returns the current value. I added that exact distinction to the worker-pool doc comment in 752b0181.

internal/cli/fmt.go
emil14: Yes, for the mode where it is useful: -w now applies changed files concurrently after the preflight in 752b0181. Print, diff, list, and check remain sequential at the output boundary because concurrent writes to their shared stdout would interleave or become scheduler-ordered; formatting itself is concurrent in every mode.

